PR #25, opened 2022-05-04
Correctly set PYTHONPATH in idpi/test/setup.sh

petrabaumann:
Reference: https://unix.stackexchange.com/questions/4650/determining-path-to-sourced-shell-script


cosunae:
launch jenkins

PR #33, opened 2022-11-19
Flexpart product

cosunae:


idpi/src/operators/flexpart.py
```diff
@@ -0,0 +1,86 @@
...
+        "V_10M",
+        "T_2M",
+        "TD_2M",
+        "CLCT",
+        "W_SNOW",
+    ):
+        ds_out[field] = ds[field].isel(step=istep)
+
+    ds_out["TOT_CON"] = (
+        (ds["TOT_CON"].isel(step=istep) - ds["TOT_CON"].isel(step=istep - 1))
```
petrabaumann, line 41: This factor accounts for the fact that FLEXPART expects precipitation rates per h. One should add a comment.

idpi/src/operators/flexpart.py
```diff
@@ -0,0 +1,86 @@
...
+    ds_out["SSHF"] = (ds["SSHF"].isel(step=istep) - ds["SSHF"].isel(step=istep - 1)) / (
+        3600 * 3
+    )
+    ds_out["SSHF"].attrs = ds["SSHF"].attrs
+    ds_out["EWSS"] = (ds["EWSS"].isel(step=istep) - ds["EWSS"].isel(step=istep - 1)) / (
+        3600 * 3
+    )
+    ds_out["EWSS"].attrs = ds["EWSS"].attrs
+
+    surface_pressure_ref = 101325.0
```
petrabaumann, line 65: One should shift all physical constants into a separate module and import that then.

idpi/test/test_flexpart.py
```diff
@@ -0,0 +1,239 @@
...
+    ds = load_data(constants + inputf, field_map, datafile)
+
+    for h in range(3, 10, 3):
+        datafile = datadir + f"/efsf00{h:02d}0000"
+        newds = load_data(inputf, field_map, datafile)
+
+        for field in newds:
+            ds[field] = xr.concat([ds[field], newds[field]], dim="step")
+
+    NV = ds["U"].GRIB_NV
```
petrabaumann, line 83: Add some comments on the steps that are implemented here.

idpi/test/test_flexpart.py
```diff
@@ -0,0 +1,239 @@
...
+
+    executable = "/project/s83c/fieldextra/tsa/bin/fieldextra_gnu_opt_omp"
+
+    # create the tmp dir
+    shutil.rmtree(tmpdir, ignore_errors=True)
+    os.mkdir(tmpdir)
+
+    templateLoader = jinja2.FileSystemLoader(searchpath=testdir + "/fe_templates")
+    templateEnv = jinja2.Environment(loader=templateLoader)
+    template = templateEnv.get_template("./test_flexpart.nl")
```
petrabaumann, line 119: I think that the template has not been checked in!

petrabaumann (changes requested):
I hink that one should add some comments on certain operations, and review the mapping between IFS and COSMO parameters.

idpi/test/test_flexpart.py
cosunae: I moved all this to a function. But reminder, I would like to discuss this with you before merge

idpi/test/field_mappings.yml
```diff
@@ -0,0 +1,122 @@
+FIS:
+  ifs:
+    name: z
+    paramId: 129
+  cosmo:
+    paramId: 500004
```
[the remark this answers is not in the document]
cosunae: ok, thanks a lot for this info. 
We need to work on a solution for this then.

idpi/test/field_mappings.yml
```diff
@@ -0,0 +1,122 @@
...
+  ifs:
+    name: tcc
+    paramId: 164
+  cosmo:
+    paramId: 500046
+W_SNOW:
+  ifs:
+    name: sd
+    paramId: 141
+  cosmo:
```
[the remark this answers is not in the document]
cosunae: Petra, there is snow depth, with units m, but snow depth water equivalent: 
https://github.com/ecmwf/eccodes/blob/b1337d386a0f5d4876e4d4b13d5ed285252d92e0/definitions/grib2/paramId.def#L7145
triplet (0,1,60) 
where units should be kg m-2, in this case there is no unit conversion needed for sd, isnt? 
Can you double check?
https://github.com/ecmwf/eccodes/blob/b1337d386a0f5d4876e4d4b13d5ed285252d92e0/definitions/grib2/units.def#L7145

idpi/test/field_mappings.yml
```diff
@@ -0,0 +1,122 @@
...
+    paramId: 142
+  cosmo:
+    paramId: 500042
+SSR:
+# IFS and cosmo not compatible
+  ifs:
+    name: ssr
+    paramId: 176
+  cosmo:
+    paramId: 500078
```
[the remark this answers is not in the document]
cosunae: I changed the key in the mapping from SSR to ASOB_S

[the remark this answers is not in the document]
cosunae: no unit conversion factor required from W to J

idpi/test/field_mappings.yml
```diff
@@ -0,0 +1,122 @@
...
+SSHF:
+# IFS and cosmo not compatible
+  ifs:
+    name: sshf
+    paramId: 146
+  cosmo:
+    paramId: 500087
```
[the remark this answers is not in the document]
cosunae: also changed SSHF to ASHFL_S

cosunae:
launch jenkins

idpi/src/operators/flexpart.py
```diff
@@ -0,0 +1,118 @@
...
+    ds["V"] = ds["V"].sel(hybrid=slice(40, 60))
+    ds["ETADOT"] = ds["ETADOT"].sel(hybrid=slice(1, 60))
+    ds["T"] = ds["T"].sel(hybrid=slice(40, 60))
+    ds["QV"] = ds["QV"].sel(hybrid=slice(40, 60))
+
+    return ds
+
+
+def append_pv(ds):
+    """Compute ak, bk (weights that define the vertical coordinate) from pv."""
```
cosunae, line 56: @petrabaumann how is fieldextra extracting ak, bk ? 
here we are assuming that ak, bk is the same for all fields, so it just takes it from U. 
It is not a good implementation since it does not even check whether U is present. But I wanted to know first how fieldextra does it before fixing it

idpi/share/field_mappings.yml
```diff
@@ -0,0 +1,133 @@
...
+    paramId: 164
+  cosmo:
+    paramId: 500046
+W_SNOW:
+  ifs:
+    name: sd
+    paramId: 228141
+  cosmo:
+    paramId: 500044
+
```
petrabaumann, line 94: It is true that FLEXPART-IFS expects SD in m, while FLEXPART-COSMO expects the equivalent parameter in kg m-2 and does the conversion to m in the import. So this is just the inverse of the transformation that we try to implement here. That is what I interpret from readwind_ecmwf.f90 in FLEXPART-IFS and readwind.f90 of FLEXPART-COSMO.
A general conversion from the IFS parameter SD to the COSMO parameter W_SNOW would involve a unit conversion from m to kg m-2. 
The import of FLEXPART must be cleaned up to achieve a clean implementation of the transformations here.

petrabaumann (commented):
This is a much improved implementation of the FLEXPART products for idpi!

idpi/src/operators/time_operators.py
```diff
@@ -0,0 +1,21 @@
+# Third-party
+import numpy as np
+import xarray as xr
+
+
+def time_rate(var: xr.DataArray, dtime: np.timedelta64):
+    coord = var.coords["step"]
+    return (
+        var.isel(step=slice(1, None))
```
petrabaumann, line 9: Note that the value of the resulting rate has to be set to undefined (np.nan) for step=0.

idpi/src/operators/flexpart.py
```diff
@@ -0,0 +1,118 @@
...
+        "V_10M",
+        "T_2M",
+        "TD_2M",
+        "CLCT",
+        "W_SNOW",
+    ):
+        ds_out[field] = ds[field].isel(step=istep)
+
+    ds_out["TOT_CON"] = time_rate(
+        ds["TOT_CON"].isel(step=slice(istep - 1, istep + 1)), np.timedelta64(1, "h")
```
petrabaumann, line 92: Here you are assuming that the temporal resolution is 1h, while in the unit test you are working with data at a temporal resolution of 3h. I suspect that the rate computation and the time incrementation are inconsistent. Could it be that precipitation is 0 in your test data set?

petrabaumann (changes requested):
I still have some comments that should be checked, in particular regarding the implementation of the rate operator.

PR #35, opened 2022-12-03
add a pyproject.toml project setup

cosunae:
Technical Description
==================

* add a pyproject.toml, so that we can install the project with pip (in editable mode)
* replace setup.cfg (by pyproject.toml)
* Modify the instructions to install and run in the README.md and in jenkinsfile

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,8 @@ Prototype for data processing framework
 conda env create --file environment_dev.yml
 conda activate idpi_env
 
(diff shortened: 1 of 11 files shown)


cosunae:
launch jenkins

cosunae:
I think we should update this project at some point to the latest blueprint, since now it is released and working. For the time being, I just copied the pyproject.toml from the blueprint by hand

PR #40, opened 2023-03-24
Flexpart run

cosunae:
Description
============
* Add a flexpart operator for pre-processing of IFS data for flexpart
* since shortnames (i.e. the keys of the xarray datasets) are different for IFS and COSMO data, we add a mapping of names between IFS and COSMO, using a unique name identifier for each field: field_mappings.yml
* Conditional loading of the grib keys depending on the product definition template number, using a yaml configuration: 
https://github.com/MeteoSwiss-APN/icon_data_processing_incubator/pull/40/files#diff-549aa0005a8efca44a9581cdb51a0944bbeda28b89c4faf871a7bcefe0a4800e

Tests
======
It comes with a test_flexpart.py that checks the output is within tolerance wrt to Fieldextra. 
I also tested manually that flexpart (Europe) runs with the output data created by flexpart_cli (up to 5 lead times) without crashes. Since there is no validation for flexpart, I could not check the validity of the output. 

Depends on
==========
#42 
https://github.com/ecmwf/cfgrib/pull/324/files ( a hack of cfgrib is in place in this PR until the bug fix is merged and published in cfgrib)


cosunae:
launch jenkins

cosunae:
launch jenkins

victoria-cherkas (commented):
Hi Carlos, this is a large PR, and I have tried to understand the best I can what is going on. I think I understand the gist of it, but in the end my most important comments come from just scanning the ouput. As you said, flexpart runs with this output but potentially the result may be wrong given the timestep/level issue for example. I think we should test the grib_output such that it contains variables at certain levels, timesteps etc, or grib_compare against some actual fx output. 

README.md:
- `conda activate idpi_env` --> `conda activate idpi_dev`
- add a comment that tests should be run on cscs since `data_dir` is hardcoded. 

Also should we be concerned about the ECCODES ERRORS on writing the ds to GRIB? 
for example
```
ECCODES ERROR   :  concept: no match for paramId=500035
ECCODES ERROR   :  concept: input handle edition=2, centre=ecmf
ECCODES ERROR   :  Please check the Parameter Database 'https://apps.ecmwf.int/codes/grib/param-db/?id=500035'
```

Also just checking the output by hand it looks like:
ETADOT is always at timestep 0. 
CLCT is missing from the output
There is an UNKNOWN variable.
U,V,T,TD at single levels are written at level 0, there are no U_10M,V_10M,T_2M,TD_2M at levels 10 and 0 respectively.
#  394 name: U                        level:     0      0        
#  395 name: V                        level:     0      0         
#  396 name: T                        level:     0      0         
#  397 name: TD

cosunae:
Hi Victoria, 

I answer to the comments in here: 
https://github.com/MeteoSwiss-APN/icon_data_processing_incubator/pull/40#pullrequestreview-1368844747

>   conda activate idpi_env --> conda activate idpi_dev
done
>    add a comment that tests should be run on cscs since data_dir is hardcoded.
yes, I added a comment. We could work on making it CSCS independent, and it would make sense. However currently the main dependency is fieldextra, since it relies on the executable installed there. 

When you ran, there was a problem with : 
`
ECCODES ERROR   :  concept: no match for paramId=500035
ECCODES ERROR   :  concept: input handle edition=2, centre=ecmf
ECCODES ERROR   :  Please check the Parameter Database 'https://apps.ecmwf.int/codes/grib/param-db/?id=500035'
`
which I could reproduce. The issue was tricky to find so I will try to explain.
In one of the comments in this PR I explain that we normally set two paths for the eccodes definitions, one for IFS and one for COSMO/ICON data, and they are not compatible. Because of that, in the test of IFS data we only set the definitions of IFS (and not cosmo):
https://github.com/MeteoSwiss-APN/icon_data_processing_incubator/pull/40/files#diff-82f994ef3fdf04023c177ce70fd6e4499422cbe378716e00e5c3fde1b258be05R200
The same variable is defined in the two definitions with different paramId, shortName, etc. 
The problem was that at some point, we ran on the input data with the definitions of COSMO as well. And the choices, like paramId, etc are then written into the index file of cfgrib. Then next time that you read the same input file, irrespective of the definitiones set in the environment, the paramId will be picked from the index file (that is in the path of the data). Therefore, from this one, the tests here were using the COSMO conventions (instead of the IFS conventions), and where failing for everyone. 
Because that is dangerous, and produces unexpected behaviour I deactivate the index files: 
https://github.com/MeteoSwiss-APN/icon_data_processing_incubator/pull/40/commits/8ca473c761eb271ad6fab1e94b9448db1866ed00

The consequence though is that we should not use this (at least w/o reactivating the index file functionality) for benchmarking. 

The output I checked more or less by hand and by running Flexpart on the produced data. 
As you suggested it would be ideal to do a systematic comparison of the output data with the fieldextra reference. However it wont work. The solution we have here is quite hacky and not general, so there are some differences in some of grib keys. Fixing that properly would take I think considerable time. And I think this is not the right approach anyhow. First we thought this could work, i.e. xarray -> grib, but giving how hacky it is, I do not think is a viable approach to do it properly and in general. Instead I think the earthkit solution is clean since it does respect the input grib metadata. 
Let discuss what to do therefore. We could also keep this in a branch, if the solution is not convincing for a merge.

cosunae:
we will not merge this, so I will close it

PR #41, opened 2023-03-24
disable concurrent builds

twicki:


PanicSheep (approved):
Now it's equivalent to the blueprint.
LGTM

PR #42, opened 2023-03-25
fix tolerances and tests

cosunae:
Description
==========
Fixes several issues with the testing in main. 
* Pin version of eccodes to 2.25.0 in conda and in the setup.sh (for the definitions used for testing). If we do not pinned the version of eccodes it could happen that the version of eccodes installed by cfgrib would be incompatible with the eccodes definitions used in the setup.sh. 
* Update some tolerances. 
* Disable the undef_fold mode of intpl_k2theta since it currently does not verify. 
* Created an issue #43  in order to recover the test that was disabled here.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -4,5 +4,6 @@ channels:
 dependencies:
 - pytest>=7.1.1
 - cfgrib>=0.9.10.1
+- eccodes==2.25.0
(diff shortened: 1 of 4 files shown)


twicki (commented):
LGTM

PR #48, opened 2023-04-25
Enabling of sphinx documentation

twicki:
## Purpose

This PR enables the functionality of automatically generating documentation provided by the blueprint.
This was not done in the initial [PR](https://github.com/MeteoSwiss-APN/icon_data_processing_incubator/pull/46) to keep the changes only to an infrastructure level and not change functionality.

The documentation was successfully built [here](https://github.com/MeteoSwiss-APN/icon_data_processing_incubator/actions/runs/4794597490/jobs/8528159018) as part of the PR but is disabled again for PR's and will only be built on commits to main - see commit [here](https://github.com/MeteoSwiss-APN/icon_data_processing_incubator/commit/3849ac3c69ee2bb49080c1bd95d477da88110661)

## Code changes:

- Following the guidelines of the blueprint [here](https://meteoswiss-apn.github.io/mch-python-blueprint/documentation.html) to enable documentation


.github/workflows/documentation.yaml
```diff
@@ -4,9 +4,6 @@ on:
   push:
     branches:
     - main
-  pull_request:
```
cosunae, line 7: yes, perhaps we should remove this from the blueprint as well

cosunae (approved):
looks good to me

.github/workflows/documentation.yaml
twicki: can be done!

PR #49, opened 2023-04-26
Added pot_vortic module

cfkanesan:
I tried to rebase the original branch but the history was too convoluted and I was afraid of introducing new bugs in the conflict resolution so I had to resort to copying the files over. I also made an attempt at using pytest fixtures to manage the fieldextra run to unload the test body a bit. Let me know what you think.

The tests are currently failing but you can already have a look and comment where relevant.

### DIFF
--- a/src/idpi/constants.py
+++ b/src/idpi/constants.py
@@ -1,0 +1,13 @@
+"""Physical constants."""
+
+pc_omega = 7.2921e-5
+earth_radius = 6371229.0  # value from COSMO GRIB2 key
+
+p0 = 1.0e5
+
+pc_r_d = 287.05
+pc_r_v = 461.51  # Gas constant for water vapour[J kg-1 K-1]
+pc_cp_d = 1005.0
+pc_rvd = pc_r_v / pc_r_d
+pc_rdocp = pc_r_d / pc_cp_d
+pc_rvd_o = pc_rvd - 1.0
--- a/src/idpi/operators/curl.py
+++ b/src/idpi/operators/curl.py
@@ -1,0 +1,136 @@
+"""Algorithm for the curl operator."""
+
+# Standard library
+from typing import cast
+
+# Third-party
+import numpy as np
+import xarray as xr
+
+# Local
+from .. import constants as const
+
+DEFAULT_DIFF_TYPE = "center"
+STENCILS = {
+    "left": {"m": slice(0, -2), "c": slice(1, -1), "p": slice(1, -1)},
+    "center": {"m": slice(0, -2), "c": slice(1, -1), "p": slice(2, None)},
+    "right": {"m": slice(1, -1), "c": slice(1, -1), "p": slice(2, None)},
+}
+
+
+def stpt(stencil: str, diff_type: str | None = None) -> dict[str, slice]:
+    """Get stencil points.
+
+    This dictionary can be used for selecting values from a data array.
+    Consider the following example, where ``A`` is a data array. ::
+
+        B = A[stpt('ccc')] + A[stpt('cpm')]
+
+    Here we add for every point a in ``A`` its upper (y+1) neighbor, one level
+    below (z-1) and store the result in ``B``.
+    In this method we allow 3d neighbor stencils. Points on the border won't
+    be included.
+
+    The stencil definition string must have three characters, corresponding to
+    the 'x', 'y' and 'z' dimensions respectively.
+    The characters can be
+    - 'c': center (+/-0)
+    - 'm': below (-1)
+    - 'p': above (+1)
+
+    Depending on the ``diff_type`` (set with ``set_diff_type``), the stencils
+    will be interpreted differently.
+    For 'left', positive stencils (+1, 'p') will be mapped to the original point.
+    For 'right', negative stencils (-1, 'm') will be mapped to the original point.
+
+    Args:
+        stencil (str): A string defining the stencil.
+        diff_type (str): A string indicating the difference type.
+
+    """
+    if diff_type is None:
+        diff_type = DEFAULT_DIFF_TYPE
+    s = STENCILS[diff_type]
+    return {d: s[k] for d, k in zip("xyz", stencil)}
+
+
+class FDField:
+    """Syntaxic sugar to apply the stencil to a field."""
+
+    def __init__(self, field: xr.DataArray, diff_type: str = DEFAULT_DIFF_TYPE):
+        """Init FDField."""
+        self.field = field
+        self.diff_type = diff_type
+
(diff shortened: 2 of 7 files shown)


src/idpi/constants.py
cfkanesan, line 1: This was present in Thierry's branch but it seems that the main has diverged from this approach. Should I drop it?

src/idpi/operators/curl.py
```diff
@@ -0,0 +1,122 @@
...
+import numpy as np
+
+from .. import constants as const
+
+DEFAULT_DIFF_TYPE = "center"
+STENCILS = {
+    "left": {"m": slice(0, -2), "c": slice(1, -1), "p": slice(1, -1)},
+    "center": {"m": slice(0, -2), "c": slice(1, -1), "p": slice(2, None)},
+    "right": {"m": slice(1, -1), "c": slice(1, -1), "p": slice(2, None)},
+}
```
cfkanesan, line 11: I modified the handling of the stencils to be more static and less dependent on the mutation of the global state. That can be a pain to work with in a concurrent execution context.

src/idpi/operators/curl.py
```diff
@@ -0,0 +1,122 @@
...
+        - 'm': below (-1)
+        - 'p': above (+1)
+    """
+    if diff_type is None:
+        diff_type = DEFAULT_DIFF_TYPE
+    s = STENCILS[diff_type]
+    return {d: s[k] for d, k in zip("xyz", stencil)}
+
+
+class FDField:
```
cfkanesan, line 45: The purpose of this class is to wrap the dataarray and provide syntaxic sugar for the expressions below.

src/idpi/operators/curl.py
```diff
@@ -0,0 +1,122 @@
...
+    wk: float,
+    sqrtg_r_s: xr.DataArray,
+    lat: xr.DataArray,
+) -> tuple[xr.DataArray, xr.DataArray, xr.DataArray]:
+    # prepare parameters
+
+    r_earth_inv = 1 / const.earth_radius
+    deg2rad = np.pi / 180
+    acrlat: xr.DataArray = 1 / (np.cos(lat * deg2rad) * const.earth_radius)  # type: ignore[assignment]
+    tgrlat: xr.DataArray = np.tan(lat * deg2rad)  # type: ignore[assignment]
```
cfkanesan, line 75: One of the issues in the original branch is that the lattitude values are in degrees and not micro degrees.

src/idpi/operators/pot_vortic.py
```diff
@@ -0,0 +1,143 @@
...
+        xr.DataArray: The potential vorticity
+    """
+    stpt = partial(stpt_, diff_type=diff_type) if diff_type else stpt_
+
+    # prepare parameters
+    dlon = U.attrs["GRIB_iDirectionIncrementInDegrees"]
+    dlat = U.attrs["GRIB_jDirectionIncrementInDegrees"]
+    deg2rad = np.pi / 180
+    inv_dlon = 1 / (dlon * deg2rad)
+    inv_dlat = 1 / (dlat * deg2rad)
```
cfkanesan, line 69: The confusion on the units was present here too

src/idpi/operators/pot_vortic.py
```diff
@@ -0,0 +1,143 @@
...
+
+    # FD (finite differences) weights
+    wi = 0.5 * inv_dlon
+    wj = 0.5 * inv_dlat
+    wk = 0.5 if diff_type == "center" else 1.0
+
+    # remove coordinates
+    U, V, W, P, T, HHL, QV, QC, QI = map(
+        lambda a: a.drop_indexes("z"), (U, V, W, P, T, HHL, QV, QC, QI)
+    )
```
cfkanesan, line 85: I'm not the biggest fan of this but at least it doesn't mutate the input variables. I still need to confirm that this works.

tests/test_idpi/test_potvortic.py
```diff
@@ -0,0 +1,61 @@
...
+    fs_ds = fieldextra("POT_VORTIC")
+    
+    pv_ref = (
+        fs_ds.rename({"x_1": "x", "y_1": "y", "z_1": "z"})
+        .squeeze(drop=True)
+        .isel(x=slice(1, -1), y=slice(1, -1))  # discard the (undefined) border
+    )
+
+    print(potv.data)
+    print(pv_ref.data)
```
cfkanesan, line 55: leftovers to remove

tests/test_idpi/test_potvortic.py
```diff
@@ -0,0 +1,61 @@
+import numpy as np
+import xarray as xr
+import pytest
+
+from idpi import grib_decoder
+import idpi.operators.pot_vortic as pv
+
+
+@pytest.fixture
+def fieldextra_executable(fieldextra_executable):
```
cfkanesan, line 10: Here is an example of the fixture being overridden at the module level.

[] remove this before merging

src/idpi/operators/curl.py
```diff
@@ -0,0 +1,136 @@
...
+    dzeta_dphi = (
+        0.25
+        * inv_dlat
+        * sqrtg_r_s
+        * ((hhl["cpc"] - hhl["cmc"]) + (hhl["cpp"] - hhl["cmp"]))
+    )
+
+    # compute weighted derivatives for FD
+
+    du_dphi = ((u["cpc"] + u["mpc"]) - (u["cmc"] + u["mmc"])) * 0.25 * inv_dlat
```
cosunae, line 109: I am not very convinced about the usage of this custom syntax. Also notice that sometimes we use it as u["cmc"]  while below we also use u("cmc"). 

I will suggest looking into more standard syntax or aligned with other projects.
* variant 1: 
```u[0,1,0]``` equals the syntax used here as ```u["cpc"]```
it is used also in numba:
https://numba.pydata.org/numba-doc/latest/user/stencil.html
and in gt4py (cartesian mode)
https://github.com/GridTools/gt4py/blob/main/examples/cartesian/demo_horizontal_diffusion.ipynb
* variant 2:
Use a shift operator (concept of gt4py)
```shift(J, 1)(u)``` is equivalent to ```u["cpc"]```
reference: https://hackmd.io/ueLehFeOR5yyOCtn2Qdr-A

* variant 3: 
Use an offset object (used in icon4py)
```u[JOff(1)]``` is equivalent ```u["cpc"]```
reference: 
https://github.com/C2SM/icon4py/blob/main/atm_dyn_iconam/src/icon4py/atm_dyn_iconam/mo_solve_nonhydro_stencil_04.py

cosunae (changes requested):

PR #52, opened 2023-05-10
Add potential vorticity operator

cfkanesan:
## Purpose

Add the potential vorticity operator

## Code changes

- add differential operators using central finite difference schemes
- add a class to encapsulate the correction terms due to terrain following coordinates
- introduce test fixtures
- add the constants module
- add the curl (rotational) operator
- add the total density operator

APNRZ-20


src/idpi/operators/stencils.py
```diff
@@ -0,0 +1,110 @@
...
+        self.padded = padded
+
+    def __getitem__(self, indices: tuple[int, int, int]) -> xr.DataArray:
+        i, j, k = indices
+        s = STENCILS["half"]
+        return self.field[{"x": s[i], "y": s[j], "z": s[k]}]
+
+    def dx(self):
+        if self.dim == "x":
+            return self[1, 0, 0] - self[-1, 0, 0]
```
cosunae, line 73: as mentioned previously I think operators of a framework should be composable and adhere to common, well described data model and syntax. Until now we were using the xarray isel by label indexing. In this PR, we combine the use of the xarray isel, a more numba style, and later in the tests bare numpy indexing. Letting this grow in a framework will lead to mix and unclear syntaxes and patterns for data access, with very different styles in various parts of the framework. I would suggest to adhere to the xarray isel until we decide to adopt another one, to avoid organic, nonuniform growth of styles.

src/idpi/operators/stencils.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class TotalDiff:
+    def __init__(self, dlon: float, dlat: float, hhl: xr.DataArray):
+        self.dlon = dlon
+        self.dlat = dlat
+        z = "generalVertical"
+        hhl_pad = PaddedField(hhl.rename({z: "z"}))
+        dh_dx = destagger_z(hhl_pad.dx())  # order is important
+        dh_dy = destagger_z(hhl_pad.dy())  # diff then destagger
```
cosunae, line 98: just a note: while this pattern of composing operators as pure functions is very readable, it will carry significant performance impact. Once this is working I think we should create a version where operator modify the inputs without creating (return) copies, i.e. have side effects, in order to compare. 
I created an issue for this: 
https://meteoswiss.atlassian.net/browse/APNRZ-102

src/idpi/operators/vertical_reduction.py
```diff
@@ -297,10 +297,12 @@ def integrate_k(field, operator, mode, height, h_bounds, hsurf=None):
     # ... note that the dimension "generalVericalLayer" is lost of this condition is
     # nowhere satisfied
     field_in_h_bounds = field_on_fl.where((hfl >= h_bottom) & (hfl <= h_top)).dropna(
-        "generalVerticalLayer"
+        dim="generalVerticalLayer",
+        how="all",
```
cosunae, line 301: this is a significant change in behaviour, and the tests were insensitive to it (since there are no changes to testing). Could we test this expected behaviour (i.e. "all" vs "any")?

cosunae (changes requested):

src/idpi/operators/curl.py
```diff
@@ -9,128 +9,45 @@
...
-        * ((hhl["pcc"] - hhl["mcc"]) + (hhl["pcp"] - hhl["mcp"]))
-    )
-
-    dzeta_dphi = (
-        0.25
-        * inv_dlat
-        * sqrtg_r_s
-        * ((hhl["cpc"] - hhl["cmc"]) + (hhl["cpp"] - hhl["cmp"]))
-    )
+    acrlat = cast(xr.DataArray, 1 / (np.cos(lat) * const.earth_radius))
```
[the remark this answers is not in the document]
cfkanesan: The type is already an xr.DataArray but mypy doesn't understand it that way from the np.cos signature. `cast` is actually a no-op for the interpreter.

src/idpi/operators/destagger.py
```diff
@@ -1,39 +1,31 @@
...
-
-        field_k1_ = field[{dim: slice(1, None)}].assign_coords(
-            {dim: field[{dim: slice(0, -1)}].generalVertical}
-        )
-        return ((field_k0_ + field_k1_) * 0.5).rename(
-            {"generalVertical": "generalVerticalLayer"}
-        )
-
-    raise RuntimeError("Unknown dimension", dim)
+        result = 0.5 * (field + field.shift({dim: -1})).dropna(dim)
```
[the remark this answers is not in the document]
cfkanesan: The staggered fields have one more element in the direction of the staggering. Without this, you need to manually slice or rely on the xarray arithmetic join to ensure the consistency of the operation.

src/idpi/operators/pot_vortic.py
```diff
@@ -78,94 +50,25 @@ def fpotvortic(
...
-        * ((hhl("cpc") - hhl("cmc")) + (hhl("cpp") - hhl("cmp")))
-    )
-
-    # curl
-    curl1, curl2, curl3 = curl(U, V, W, HHL, inv_dlon, inv_dlat, wk, sqrtg_r_s, lat)
-
-    # potential temperature
-    theta = ftheta(P, T)
+    deg2rad = np.pi / 180
+    lat = (rho_tot["latitude"] * deg2rad).astype(np.float32)
```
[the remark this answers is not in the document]
cfkanesan: The latitude coordinates are stored in float64

src/idpi/operators/stencils.py
```diff
@@ -0,0 +1,110 @@
...
+        return 0.5 * (self[0, 1, 0] - self[0, -1, 0])
+
+    def dz(self) -> xr.DataArray:
+        return 0.5 * (self[0, 0, 1] - self[0, 0, -1])
+
+
+class StaggeredField:
+    """Should have one more element along the staggered dimension."""
+
+    def __init__(self, field: xr.DataArray, padded: PaddedField, dim: Dim):
```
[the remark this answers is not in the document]
cfkanesan: I agree that the signature is not the best but I chose to make the sacrifice for performance reasons. The contents of the padded field and the staggered field are not the same. The padded field contains the values at the destaggered positions whereas the `field` contains values at the staggered positions. I could destagger within the function call but this is a relatively expensive operation.

src/idpi/operators/pot_vortic.py
```diff
@@ -78,94 +50,25 @@ def fpotvortic(
...
 
     # coriolis terms
-    cor2 = 2 * const.pc_omega / const.earth_radius * np.cos(lat * deg2rad)
-    cor3 = 2 * const.pc_omega * np.sin(lat * deg2rad)
+    cor2 = 2 * const.pc_omega / const.earth_radius * np.cos(lat)
+    cor3 = 2 * const.pc_omega * np.sin(lat)
 
-    # total air density
-    rho = f_rho_tot(T, P, QV, QC, QI, QW_load)
+    t = PaddedField(theta.rename(generalVerticalLayer="z"))
```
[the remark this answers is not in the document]
cfkanesan: This is a fair point, I can try to stick to general vertical and general vertical layer for as much as possible.

src/idpi/operators/vertical_reduction.py
cfkanesan: Sure. If a single column is undefined, then the entire output is empty. Should I add this as a test case?

src/idpi/operators/stencils.py
cfkanesan: I would argue that when you have a mix of pure and impure functions, everybody starts making defensive copying "just in case".

src/idpi/operators/stencils.py
cfkanesan: Can you provide some additional detail on how to achieve this?

src/idpi/operators/stencils.py
```diff
@@ -0,0 +1,115 @@
...
+        dh_dy = destagger_z(hhl_pad.dy())  # diff then destagger
+
+        dim = "generalVertical"
+        rename_kw = {dim: "generalVerticalLayer"}
+        self.sqrtg_r_s = -1 / hhl.diff(dim=dim, label="lower").rename(rename_kw)
+        self.dzeta_dlam = self.sqrtg_r_s / dlon * dh_dx
+        self.dzeta_dphi = self.sqrtg_r_s / dlat * dh_dy
+
+    def d_dlam(self, field: Field) -> xr.DataArray:
+        return field.dx() / self.dlon + field.dz() * self.dzeta_dlam
```
cosunae, line 109: I am not sure I understand why dx, dy were coded different than dz. 
I think the delta function is `(self[1, 0, 0] - self[-1, 0, 0])` irrespective of the axis, instead of `0.5 * (self[1, 0, 0] - self[-1, 0, 0])`. 
In the the x dimension here we would have: 
`delta(field, 'x') / (2*self.dlon)`
very similar for the z dimension but instead of 2*(hhl[0,0,1]-hhl[0,0,0], we need to use the geometrical coefficient dzeta_dlam. So the partial derivatives are different, but this is due to the differences between z and horizontal dimension, and not related to staggering, isnt?
So like that the delta function can be coded invariant wrt the axis.

cfkanesan: The idea was to make `dx, dy, dz` return the derivatives of the field with respect to unit length coordinates in the 3 directions respectively. Then, multiply those by the grid deformation factors. I am aware that there is some inconsistency on the naming but I tried to stick to the naming in fieldextra. I'm a bit unclear on the naming conventions so perhaps I have introduced some further inconsistencies. Also, note that the `self[1, 0, 0]` returns different stencils whether you're looking at `PaddedField` or `StaggeredField` and the first has a different treatment of the vertical as well. This is done to maintain consistency with the vertical boundary management in fieldextra. I agree that these are all potential sources of confusion but I fail to see how changing the unit length to an arbitrary length will help reduce the confusion.

src/idpi/operators/pot_vortic.py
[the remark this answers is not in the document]
cosunae: ok, thanks. Probably we should control that at loading time, and make sure everything comes as float32 (unless there is a reason not to)?, but anyhow not for this PR

src/idpi/operators/pot_vortic.py
```diff
@@ -78,94 +52,25 @@ def fpotvortic(
...
 
     # coriolis terms
-    cor2 = 2 * const.pc_omega / const.earth_radius * np.cos(lat * deg2rad)
-    cor3 = 2 * const.pc_omega * np.sin(lat * deg2rad)
+    cor2 = 2 * const.pc_omega / const.earth_radius * np.cos(lat)
+    cor3 = 2 * const.pc_omega * np.sin(lat)
 
-    # total air density
-    rho = f_rho_tot(T, P, QV, QC, QI, QW_load)
+    dt_dlam = total_diff.d_dlam(diff.dx(theta), diff.dz(theta))
```
cosunae, line 67: question about the naming: what is the meaning of total_diff (or class TotalDiff)?

src/idpi/operators/diff.py
```diff
@@ -0,0 +1,43 @@
...
+
+def dy(field: xr.DataArray) -> xr.DataArray:
+    return field.copy(data=_dxy(field, axis=-2))
+
+
+def dz(field: xr.DataArray) -> xr.DataArray:
+    return field.copy(data=np.gradient(field, axis=-3))
+
+
+def dz_staggered(field: xr.DataArray) -> xr.DataArray:
```
cosunae, line 41: it would be good to add some documentation, particularly in the differences between the x,y and z and this last one that is directly changing the staggered location

src/idpi/operators/pot_vortic.py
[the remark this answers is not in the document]
cfkanesan: I think it's documented in fieldextra that coordinates are always stored in double precision but I don't know why

src/idpi/operators/pot_vortic.py
cfkanesan: We discussed this last week and I was asking if you had any ideas for better naming. The class provides operators that take into account the additional terms due to the terrain following grid deformation. It is however lacking the terms due to the curvature of the earth so total diff is not the most appropriate term.

src/idpi/operators/diff.py
cfkanesan: In theory, there should be staggered variants in the x and y directions too. However, due to the missing first element, I'm not sure how that boundary should be treated. The curl operator does not require the `du_dx` and `dv_dy` terms so I don't think we can rely on that specific implementation in fieldextra. I think it's wiser to implement them when we come across an actual use case.

src/idpi/operators/pot_vortic.py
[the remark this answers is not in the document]
cosunae: interesting, can we ask @petrabaumann @jmbettems why is that?

cfkanesan:
Rebased on main and squashed some redundant commits

src/idpi/operators/vertical_reduction.py
cosunae: I do not know. I think the proposed change here is correct since we do not want to drop entire column because there is any nan, I just do not understand how this was working before and passing tests. But also it seems that this algorithm is reducing the horizontal domain of the resulting field, instead of returning a field with values nan in horizontal grid points where the hfl within [h_bottom, h_top] condition is not satisfied. So I do not understand why we need a dropna here, particularly when there is a `fillna(0.0)` later in the algorithm. we could briefly discuss it to understand it, but anyhow I think it does not belong to this PR

tests/test_idpi/test_diff.py
```diff
@@ -0,0 +1,43 @@
...
+import numpy as np
+from numpy.testing import assert_allclose
+
+# First-party
+from idpi import grib_decoder
+from idpi.operators import diff
+from idpi.operators.theta import ftheta
+
+
+def test_padded_field(data_dir, grib_defs):
```
cosunae, line 11: there is not anymore any padded field in this PR. I guess the name is based on the previous implementation, so should we rename the test?

tests/test_idpi/conftest.py
```diff
@@ -0,0 +1,72 @@
...
+        conf_files = {
+            "inputi": data_dir / "lfff<DDHH>0000.ch",
+            "inputc": data_dir / "lfff00000000c.ch",
+            "output": f"<HH>_{field_name}.nc",
+        }
+        template = template_env.get_template(f"test_{field_name}.nl")
+        nl_path = tmp_path / f"test_{field_name}.nl"
+        nl_path.write_text(template.render(file=conf_files))
+
+        print(fieldextra_executable)
```
cosunae, line 49: Is this print intended?

tests/test_idpi/test_potvortic.py
```diff
@@ -0,0 +1,44 @@
...
+
+    dlon = ds["HHL"].attrs["GRIB_iDirectionIncrementInDegrees"]
+    dlat = ds["HHL"].attrs["GRIB_jDirectionIncrementInDegrees"]
+    deg2rad = np.pi / 180
+
+    total_diff = TotalDiff(dlon * deg2rad, dlat * deg2rad, ds["HHL"])
+
+    observed = pv.fpotvortic(ds["U"], ds["V"], ds["W"], theta, rho_tot, total_diff)
+
+    fs_ds = fieldextra("POT_VORTIC")
```
cosunae, line 34: I like this change, but I think we should have consistent patterns for all other integration tests that run fieldextra namelists. Can you propagate this?

cosunae:
Also please modify the description of the PR, since it seems to refer to an old implementation

src/idpi/operators/diff.py
```diff
@@ -0,0 +1,121 @@
...
+
+    """
+    return field.copy(data=_dxy(field, axis=-2))
+
+
+def dz(field: xr.DataArray) -> xr.DataArray:
+    """Compute central finite difference along z-axis.
+
+    The z-axis is assumed to be the third to last dimension of the array.
+    The coordinates spacing is assumed to be of unit length.
```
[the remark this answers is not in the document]
cfkanesan: These operators are written under the assumption that the grid is on a regular Cartesian grid. The fact that the physical grid has a `dlon, dlat` spacing in `x, y` resp. comes in the total diff class with the `1/dlon` and `1/dlat` factors.

src/idpi/operators/pot_vortic.py
```diff
@@ -0,0 +1,79 @@
...
+    Returns
+    -------
+    xr.DataArray:
+        The potential vorticity
+
+    """
+    # target coordinates
+    deg2rad = np.pi / 180
+    lat = (rho_tot["latitude"] * deg2rad).astype(np.float32)
+    rlat = get_rotated_latitude(w)
```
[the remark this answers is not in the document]
cfkanesan: It's a 1D that is relatively cheap to compute on the spot. I believe we would need to profile performance and memory use to figure out whether it's better to cache or to recompute.

tests/test_idpi/conftest.py
cfkanesan: Probably a leftover

tests/test_idpi/test_potvortic.py
cfkanesan: Can I do this in another PR?

src/idpi/operators/vertical_reduction.py
cfkanesan: I believe the intention is drop the levels that do not intersect with h bounds in order to reduce the number summations of 0 values. I have had to change the `fill(0.0)` line in the next as well because it did not agree with the way the fieldextra is handling some of these undefined values.

tests/test_idpi/test_potvortic.py
cfkanesan: I went ahead and made the PR: https://github.com/MeteoSwiss-APN/icon_data_processing_incubator/pull/55

cosunae:
thanks looking into the reviews. This looks good to me. If you modify the description (at least I think this sentence `padding the borders to handle the edge of the domain` does not apply anymore?) we can approve this

src/idpi/operators/vertical_reduction.py
cfkanesan: note to self: add comment here

PR #53, opened 2023-05-24
Destagger ufunc

cfkanesan:
## Purpose

Update the destagger operator to improve the performance and drop the coordinates in the horizontal cases since those are stale. 

## Code changes:

- Use `apply_ufunc` to improve the performance
- Drop the coordinates in the horizontal cases

## Review

### DIFF
--- a/src/idpi/operators/destagger.py
+++ b/src/idpi/operators/destagger.py
@@ -1,6 +1,17 @@
 """algorithm for destaggering a field."""
 # Third-party
+import numpy as np
 import xarray as xr
+
+
+def _dstgr(a: np.ndarray):
+    t = a.copy()
+    t[..., 1:] = 0.5 * (a[..., :-1] + a[..., 1:])
+    return t
+
+
+def _dstgr_z(a: np.ndarray):
+    return 0.5 * (a[..., :-1] + a[..., 1:])
 
 
 def destagger(field: xr.DataArray, dim: str) -> xr.DataArray:
@@ -18,22 +29,26 @@ def destagger(field: xr.DataArray, dim: str) -> xr.DataArray:
             {"x","y","generalVerticalLayer"}
 
     """
+    dims = list(field.sizes.keys())
     if dim == "x" or dim == "y":
-        field_ = field
-        field_[{dim: slice(1, None)}] = (
-            field[{dim: slice(0, -1)}] + field[{dim: slice(1, None)}]
-        ) * 0.5
-        return field_
+        return xr.apply_ufunc(
+            _dstgr,
+            field.reset_coords(drop=True),
+            input_core_dims=[[dim]],
+            output_core_dims=[[dim]],
+        ).transpose(*dims)
     elif dim == "generalVertical":
-        field_k0_ = field[{dim: slice(0, -1)}].assign_coords(
-            {dim: field[{dim: slice(0, -1)}].generalVertical}
-        )
-
-        field_k1_ = field[{dim: slice(1, None)}].assign_coords(
-            {dim: field[{dim: slice(0, -1)}].generalVertical}
-        )
-        return ((field_k0_ + field_k1_) * 0.5).rename(
-            {"generalVertical": "generalVerticalLayer"}
+        return (
+            xr.apply_ufunc(
+                _dstgr_z,
+                field,
+                input_core_dims=[[dim]],
+                output_core_dims=[[dim]],
+                exclude_dims={dim},
+            )
+            .transpose(*dims)
+            .assign_coords({dim: field.generalVertical[:-1]})
+            .rename({"generalVertical": "generalVerticalLayer"})
         )
 
     raise RuntimeError("Unknown dimension", dim)
--- a/src/idpi/operators/destagger.py
+++ b/src/idpi/operators/destagger.py
@@ -1,4 +1,7 @@
 """algorithm for destaggering a field."""
+# Standard library
+from typing import Literal
+
 # Third-party
 import numpy as np
 import xarray as xr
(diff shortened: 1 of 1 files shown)


src/idpi/operators/destagger.py
```diff
@@ -1,39 +1,68 @@
...
+# Standard library
+from typing import Literal
+
 # Third-party
+import numpy as np
 import xarray as xr
 
 
-def destagger(field: xr.DataArray, dim: str) -> xr.DataArray:
+def _dstgr(a: np.ndarray):
```
cosunae, line 10: Some further thoughts here: this operator assumes we are destaggering data that is on edges into cell centers, because it turns out in our data model that U is shifted to the + direction by dlon/2 wrt to T (and they both have the same number of indices). We should probably check from the metadata (once we have a proper metadata defined). One could use here the coordinates but it would not be very efficient, since the coordinates are 2D arrays, while we know that all latitudes should be shifted the same way (dlon/2). The checks would for example make sure we do not try to apply this function to a T field, since they de-stagger operator should be defined differently, i.e. retaining from the original field the last element, instead of element 0. I think we could add those protections later. 

Additionally, in order to understand why the _dstgr and _dstgr_z are coded that way (and why they are different) one needs to have a mental map of how the fields are layout and staggered within the grid. 
I was thinking whether we could have a single function, where the api describes the behaviour, by adding one argument, for example `boundary='left'|'right'|None`  which defines the strategy for the boundary depending on whether we are generating a shifted field to the right or left or just shrinking it. 
Something like the picture shows
path42

cosunae: Like that:
* `boundary='left'` should be used for a destaggering of U into mass point. 
* `boundary='right'` should be used for a destaggering of T into a U point. 
* `boundary=None` should be used for a destaggering of W into a mass point. (i.e. shrinking the domain size into the center)

And as said later on we could add corresponding checks.

src/idpi/operators/destagger.py
```diff
@@ -1,39 +1,68 @@
...
+    t = a.copy()
+    t[..., 1:] = 0.5 * (a[..., :-1] + a[..., 1:])
+    return t
+
+
+def _dstgr_z(a: np.ndarray):
+    return 0.5 * (a[..., :-1] + a[..., 1:])
```
cosunae, line 17: I am not sure about this comment, but here we are supposedly passing only 1D arrays, i.e. the only dimension that has not been broadcasted from the apply_ufunc ? that means, do we still need/want the `...` ?

cfkanesan: I had the same thought but I checked and the array is just transposed.

src/idpi/operators/destagger.py
cfkanesan: Yes it's true that this is relying on the assumption that the U field is staggered in x+ but, until we have metadata to describe the staggering of a field, I don't see how this check could be implemented. Right now, it's the caller's responsibility to call the correct function same as for the original code.
Here your illustration for `boundary=None` is the same as for `boundary='right'` but I assume that you meant something more like `boundary='left'` just with an additional layer to the left.
First, I think the pattern where a function does different things based on an enum arg really means that there should be separate functions.  Second, I don't know of any case where `boundary='right'` would be used. I would suggest to delay the implementation until we have such a case. 
Note that `_dstgr` is prefixed with an underscore indicating that it is a private function that is not intended to be use outside of this module.

src/idpi/operators/destagger.py
```diff
@@ -1,39 +1,68 @@
...
-            {dim: field[{dim: slice(0, -1)}].generalVertical}
-        )
-
-        field_k1_ = field[{dim: slice(1, None)}].assign_coords(
-            {dim: field[{dim: slice(0, -1)}].generalVertical}
-        )
-        return ((field_k0_ + field_k1_) * 0.5).rename(
-            {"generalVertical": "generalVerticalLayer"}
+        return (
+            xr.apply_ufunc(
```
cosunae, line 56: can you remind me why is apply_ufunc needed? According to the doc this seems to be used for applying vectorize ufunc functions that are not xarray aware, like no.<OP>. but this is not the case here. 
If the idea is to bypass the alignment and coordinate operation, simply the removal of the coordinates by itself will already have that effect, isnt? What is the benefit of using apply_ufunct?

cfkanesan: It's a simple wrapper that allows you to work with pure numpy arrays in the ufunc. It takes care of the data alignment and transposing the array such that the dimension that you want to work along is the last one.

src/idpi/operators/destagger.py
cosunae: Yes, the last picture was wrong, here again 
path42

> First, I think the pattern where a function does different things based on an enum arg really means that there should be separate functions. 
I am not sure about this. The function does something well specified, it changes the staggered location of a field. The behaviour can be tuned at the boundaries, still that does not make it a different thing like avg or gradient are. 

> Second, I don't know of any case where boundary='right' would be used. I would suggest to delay the implementation until we have such a case.

I also do not know if we will need it, but the comment was to improve readability in terms of what the function does, not necessarily generality. I agree that we do not need to implement all options. But I argue that the amount of options that we the algorithm might generate (different tuning of the boundaries) is large and assigining different function names with arbitrary name does not help. `_dstgr_z` and `dstgr` do not say what they do, why are there different behaviour needed, etc. Unless you know well the single context where this is applied, i.e. to destagger W point into mass point. But a change of grid location from mass point to W point is also possible. 

```
def destagger(field, dim):
    if dim == "x" or dim=="y":
        return grid_reloc(field, bc="l")
    elif dim =="generalVertical":
        return grid_reloc(field, bc=None)
   elif dim == "generalVerticalLayer": 
       raise RuntimeError("I can not destagger something that is already at the mass point")

def stagger(field, dim):
    if dim == "x" or dim=="y":
        return grid_reloc(field, bc="r")
    elif dim =="generalVerticalLayer":
        return grid_reloc(field, bc="lr")
   elif dim == "generalVertical": 
       raise RuntimeError("I can not stagger something that is already staggered")

```
I find this more readable because the calling function describe with the argument the behaviour of the numerical algorithm that is being applied, and I can reason whether the dispatching is done correctly for each of the metadata options. 

Also I would opt for non private functions. By making them public, we can reuse them, but also they can be made well described algorithms, why should they be private? Making numerical algorithoms public has the advantage that forces us to have a good api, thinking about the architecture of the operators and so on.

src/idpi/operators/destagger.py
cfkanesan: Using the apply_ufunc approach here improves the performance by a factor 6 on the test data. That alone is enough reason to use it in my opinion.

src/idpi/operators/destagger.py
cfkanesan: destagger is the public interface and that is the one that users should use. I do not think it's a good practice to expose all of the internals because it becomes part of the public interface (or API) and that leads to excess coupling which then makes the code base more difficult to refactor and thus maintain. This is due to the fact that breaking changes of the API require coordination to carry out. I agree that it's important to think about the API but limiting its surface is equally important.

src/idpi/operators/destagger.py
cfkanesan: As discussed, the improvement was 6x in the xy directions (compared to the patched version that does a copy) and more like 10x in the z direction.

cosunae (commented):
I think the last iteration improved considerably the readability, thanks for the effort!
The pre-commit is failing, could you fix it so that we can merge?

cosunae (approved):

PR #54, opened 2023-05-30
Add product ninjo k2th

cfkanesan:
## Purpose
Implement the product number 2 as defined in the RZ+ project pages. The product computes the mean potential vorticity between a pair of given pressure levels and the pressure, horizontal wind components and potential vorticity at given potential temperature levels.

## Code changes:
- Added the product as described above.

APNRZ-95


src/idpi/operators/vertical_reduction.py
```diff
@@ -317,8 +317,8 @@ def integrate_k(field, operator, mode, height, h_bounds, hsurf=None):
     if "generalVerticalLayer" in field_in_h_bounds.dims:
         rfield = (
             (field_in_h_bounds * dh_in_h_bounds)
-            .fillna(0.0)
             .sum(dim="generalVerticalLayer")
+            .where(~field_in_h_bounds.isnull().all(dim="generalVerticalLayer"))
```
cosunae, line 321: is the order here correct? Isnt the idea to eliminate the nan before sum ?

src/idpi/products/pot_vortic.py
```diff
@@ -0,0 +1,68 @@
...
+def _compute_at_theta(theta, hfl, **fields):
+    logger.info(f"Interpolating {tuple(fields.keys())} at isotherms")
+    theta_values = [310, 315, 320, 325, 330, 335]
+    return {
+        key: interpolate_k2theta(field, "low_fold", theta, theta_values, "K", hfl)
+        for key, field in fields.items()
+    }
+
+
+def product2(U, V, W, T, P, QV, QC, QI, HHL):
```
cosunae, line 51: we should check with @petrabaumann if there is already an identifier/name proposed for the product, product2 is arbitrary

tests/test_idpi/test_potvortic.py
```diff
@@ -31,7 +31,7 @@ def test_pv(data_dir, fieldextra, grib_defs):
 
     observed = pv.fpotvortic(ds["U"], ds["V"], ds["W"], theta, rho_tot, total_diff)
 
-    fs_ds = fieldextra("POT_VORTIC")
+    fs_ds = fieldextra("POT_VORTIC")["POT_VORTIC"]
```
cosunae, line 34: `fs_ds` is not a dataset here anymore, I would rename to `fx_pot_vortic`

src/idpi/products/pot_vortic.py
```diff
@@ -0,0 +1,68 @@
...
+from idpi.operators.total_diff import TotalDiff
+from idpi.operators.vertical_interpolation import interpolate_k2p
+from idpi.operators.vertical_interpolation import interpolate_k2theta
+from idpi.operators.vertical_reduction import integrate_k
+
+logging.basicConfig(stream=sys.stdout, level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def _compute_pot_vortic(U, V, W, T, P, QV, QC, QI, HHL, theta):
```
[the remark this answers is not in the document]
cfkanesan: The goal of this function is to restrict the lifetime of `rho_tot` and `total_diff` to minimise the memory footprint, there's no guarantee that it will remain once we've moved to `dask.delayed` or some other computation framework.

src/idpi/products/pot_vortic.py
```diff
@@ -0,0 +1,68 @@
...
+    logger.info("Computing total density")
```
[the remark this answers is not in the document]
cfkanesan: I'm lacking a global view of the project but since this is the only product, this felt like an appropriate granularity. The log events can be demoted to debug when appropriate.

src/idpi/products/pot_vortic.py
cfkanesan: I didn't have any other reference to use for a name

src/idpi/operators/vertical_reduction.py
cfkanesan: The idea is to keep full columns of nans as nans in the output field. It seems that fieldextra is doing it this way but I haven't confirmed this.

src/idpi/products/pot_vortic.py
cosunae: I just checked with Petra. In the future we will have identifier for products that are disseminated through FabIO, but we do not have it yet. 
Petra suggested to get a name from the output filename of the Fieldextra product: 
https://github.com/MeteoSwiss-APN/oprconfig/blob/1661ea7595c257d28452e2093ff033844b1b40bb/nl_rendering/templates/c2e_fcst/products_c/ninjo_5/alpk2th_ctrl.nl#L39
In this case could be something like
```ninjo_cosmo_2e_ctrl_k2th```

Notice the filename contains `alpk2th`, which I guess refers to alp domain, which happens throught the index selection
https://github.com/MeteoSwiss-APN/oprconfig/blob/1661ea7595c257d28452e2093ff033844b1b40bb/nl_rendering/templates/c2e_fcst/products_c/ninjo_5/alpk2th_ctrl.nl#LL10C1-L11C1
But this product does not seem to implement that region selection, so that is why I woud not use  'alp' in the name

File: src/idpi/operators/vertical_reduction.py (hunk 1 above)
Comment (petrabaumann) [reply src/idpi/operators/vertical_reduction.py:321]: Yes, fieldextra keeps full columns of nans as nans in the output field. It is left to the user to reset nan values by specific values, which he can do by specifying poper="replace_undef,<value>" in the namelist (<value> to be replaced by a real number).

File: src/idpi/products/pot_vortic.py [hunk 6]
```diff
@@ -0,0 +1,68 @@
...
+    rho_tot = f_rho_tot(T, P, QV, QC, QI)
+
+    logger.info("Computing terrain following grid deformation factors")
+    dlon = HHL.attrs["GRIB_iDirectionIncrementInDegrees"]
+    dlat = HHL.attrs["GRIB_jDirectionIncrementInDegrees"]
+    deg2rad = np.pi / 180
+    total_diff = TotalDiff(dlon * deg2rad, dlat * deg2rad, HHL)
```
petrabaumann, line 30: Note that in our production the post-processing is running in parallel to the model integration, so the potential vorticity operator will be executed each time when a new model output becomes available, and not only at the point when data is available up to the maximum lead time. It should probably be avoided that total_diff will be recomputed every time when a new time step of the model output becomes available for processing. fieldextra actually uses a caching mechanism to avoid re-computation of static information.

src/idpi/products/pot_vortic.py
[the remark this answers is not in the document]
petrabaumann: I think that the question of how the API should look like still needs to be discussed.

petrabaumann (commented):
I am glad to see the first product implemented.

src/idpi/products/pot_vortic.py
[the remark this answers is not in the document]
cosunae: yes, that is fine. But if this is not a public operator, the next product (there is more than one) that needs to compute pot_vortic will have to implement a copy of this private function.

src/idpi/products/pot_vortic.py
cosunae: I agree with that. There is no need to recompute constant data every lead time. Also memory footprint should not be a concern for constant fields.

src/idpi/products/pot_vortic.py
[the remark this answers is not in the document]
cosunae: As pointed out here: https://github.com/MeteoSwiss-APN/icon_data_processing_incubator/pull/54/files#r1214583863
the main pattern will be a stream of data in the time dimension lead time by lead time. 
Whatever we do, I think we should have the same pattern and logging for all operators. I think it would make sense to create a design for it so that all operator log in the same manner, like using a decorator that will register any operator in the logging. I would do that in a separate design and PR

src/idpi/products/pot_vortic.py
cfkanesan: I think that while this is very valuable insight, implementing a caching mechanism to reuse the terrain following terms is out of scope for this PR. Let me open a ticket for this.

src/idpi/products/pot_vortic.py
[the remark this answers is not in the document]
cfkanesan: The API to compute the potential vorticity is the one that is found in the operators. I wouldn't multiply the number of ways to compute the same thing.

src/idpi/products/pot_vortic.py
cosunae: I think the domain should be an input argument (part of the API) to the operator. That way we can have this product configured on any domain. A domain is a subset of the entire hypercube, including all dimensions like ensembles. 
So I think `ctrl` is also part of a domain specification `{ens=0}`
Therefore I would reduce the name to `ninjo_k2th`

cfkanesan: Okay renamed the symbols and files

src/idpi/products/pot_vortic.py
cfkanesan: https://meteoswiss.atlassian.net/browse/APNRZ-111

src/idpi/operators/vertical_reduction.py
cfkanesan: I added a comment

src/idpi/products/pot_vortic.py
petrabaumann: Yes, agree with you, Christian. And we have not yet implemented any time looping nor time operators so far. So temporal aspects should definitely be tacked in separate PRs.

petrabaumann (approved):

petrabaumann:
I approve this PR, because questions regarding temporal aspects of processing and the notion of the API should be tackled outside.

PR #55, opened 2023-06-01
Propagate test fixtures

cfkanesan:
## Purpose

Propagate the use of the test fixtures to all the tests.

## Code changes:

- Use the test fixtures to reduce the presence of unrelated code in the test bodies


victoria-cherkas (approved):
Looks great. No comments.

PR #57, opened 2023-06-08
Refactor vertical interpolation

cfkanesan:
## Purpose

Prepare the upcoming changes related to the move to earthkit data and the new metamodel design. 

## Code changes:

- Introduce the TargetCoordinates class to describe the target vertical coordinates. This is done to avoid the use of dict typing which is a form of code smell.
- Simplify the dimensions handling in `init_field_with_vcoord`
- Added some missing type hints, use Literal to indicate the possible choices for string parameters

### DIFF
--- a/src/idpi/operators/support_operators.py
+++ b/src/idpi/operators/support_operators.py
@@ -2,19 +2,43 @@
 
 
 # Standard library
+import dataclasses as dc
 from typing import Any
-from typing import Callable
-from typing import Mapping
 from typing import Optional
+from typing import Literal
+from typing import Sequence
 
 # Third-party
 import numpy as np
 import xarray as xr
 
 
+@dc.dataclass
+class TargetCoordinatesAttrs:
+    """Attributes to the target coordinates."""
+
+    standard_name: str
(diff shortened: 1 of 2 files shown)


src/idpi/operators/support_operators.py
```diff
@@ -2,19 +2,43 @@
...
 
 
+@dc.dataclass
+class TargetCoordinatesAttrs:
+    """Attributes to the target coordinates."""
+
+    standard_name: str
+    long_name: str
+    units: str
+    positive: Literal["up", "down"]
```
twicki, line 23: is there a  specific reason to not make this an enum?

twicki (approved):
much cleaner now, LGTM

src/idpi/operators/support_operators.py
cfkanesan: I chose the string literals as a trade off. Enums are stricter at runtime (provided you respect the type hint) but they require the caller to import an additional symbol at the call site. With the string literals, the syntax is lighter but some assistance is still provided by the type checker.

PR #56, opened 2023-06-04
implement data model for interp k2p

cosunae:
## Purpose

The purpose is to show a demo of a metadata model for the internal xarray objects that are passed to (and returned from) operators. 
This is a draft with the intention to discuss the proposal. Not ready to review for a merge. 
The proposal follows Variant 3 of the design: 
https://hackmd.io/559zJIyMQSu_Qi2Sw81WRA

The metadata model proposed follows: 

### Dims
* ```dims: ('x','y',z','eps','time')```
vertical dimension is always z irrespective of whether it is based on model levels, pressure or theta
### Coordinates:
We only used coordinates for vertical coordinates system that do not form a regular grid, like pressure, theta or z, where the values are defined as a 1D array.
For example, for an array defined on 1000 Pa and 10000 Pa we use:

* ```coords={'z': [1000,10000]}```
The units of the coordinates are based on a standard unit for each of the vcoord_type. This avoids having to store additionally the metadata for the unit.
### Attributes
* ```vcoord_type: ml | z | pressure | theta```
defines the type of vertical coordinate system. Notice the type of vertical coordinate (opposite to the dimension name used in cfgrib) does not define the location and whether the field is staggered. This is provided by a separate attribute origin.
* ```origin: {'x': 0|0.5, 'y': 0|0.5, 'z': -0.5|0}```
defines the location of the origin with respect to a reference (T or mass point). Some examples are:
```U:origin: {'x': 0.5, 'y': 0, 'z': 0}```
```V:origin: {'x': 0, 'y': 0.5, 'z': 0}```
```HHL: origin: {'x': 0, 'y': 0, 'z': -0.5}```
* name of parameter
* units
```

Remove the sections below which do not apply.

## Code changes:

- This PR shows the demo on only two operators: destagger and vertical_interpolation (to pressure). Also the corresponding tests are adapted.
- earthkit is shown to load xarray for fields from files and fill the proposed metadata from the earthkit object.

## Requirements changes:

- It should modify environment to add earhkit-data (not done yet)

## Infrastructure changes:

## Checklist
Before submitting this PR, please make sure:

- [ ] You have followed the coding standards guidelines established at [Code Review Checklist](https://meteoswiss.atlassian.net/wiki/spaces/UA/pages/20157433/Factory).
- [ ] Docstrings and type hints are added to new and updated routines, as appropriate
- [ ] All relevant documentation has been updated or added (e.g. README)
- [ ] Unit tests are added or updated for non-operator code
- [ ] New operators are properly tested

[...]

PR #59, opened 2023-06-13
Use earthkit.data

cfkanesan:
## Purpose

Use the `earthkit.data` module.

Note that metadata has changed but it's not an implementation of the proposal for the staggering information. That would be the subject of another PR.

## Code changes:

- The vertical coordinate is dropped but the type of level is used as the vertical dimension. This is done to limit the side effects in the rest of the code.
- The signature of `grib_decoder.load_data` is changed to return a dict rather than mutate the argument.
- A context manager is provided to mutate the grib definitions path. Note that this is *not* thread safe at the moment.
- One of the fields must be indicated as a reference for the horizontal coordinates. All fields are labelled as being defined on those coordinates regardless of staggering. This shortcoming will be addressed in a follow up PR.

## Requirements changes:

- Add earthkit-data
- Remove cfgrib
- Remove netcdf (not a direct dependency)
- Add xarray (is a direct dependency)

### DIFF
--- a/src/idpi/grib_decoder.py
+++ b/src/idpi/grib_decoder.py
@@ -32,6 +32,22 @@ def cosmo_grib_defs():
         eccodes.codes_set_definitions_path(restore)
 
 
+def _is_ensemble(field) -> bool:
+    try:
+        return field.metadata("typeOfEnsembleForecast") == 192
+    except KeyError:
+        return False
+
+
+def _gather_coords(field_map, dims):
+    coord_values = zip(*field_map)
+    unique = (sorted(set(values)) for values in coord_values)
+    coords = {dim: c for dim, c in zip(dims[:-2], unique)}
+    ny, nx = next(iter(field_map.values())).shape
+    shape = tuple(len(v) for v in coords.values()) + (ny, nx)
+    return coords, shape
+
+
 def load_data(
     params: list[str], datafiles: list[Path], ref_param: str
 ) -> dict[str, xr.DataArray]:
@@ -66,15 +82,20 @@ def load_data(
 
     hcoords = None
     metadata = {}
-    level_types = {}
-    data: dict[str, dict[tuple[int, int], np.ndarray]] = {}
+    dims: dict[str, tuple[str, ...]] = {}
+    data: dict[str, dict[tuple[int, ...], np.ndarray]] = {}
     for field in fs.sel(param=params):
         param = field.metadata("param")
         field_map = data.setdefault(param, {})
-        field_map[field.metadata("step", "level")] = field.to_numpy(dtype=np.float32)
+        dim_keys = (
+            ("perturbationNumber", "step", "level")
+            if _is_ensemble(field)
+            else ("step", "level")
+        )
+        field_map[field.metadata(*dim_keys)] = field.to_numpy(dtype=np.float32)
 
-        if param not in level_types:
-            level_types[param] = field.metadata("typeOfLevel")
+        if param not in dims:
+            dims[param] = dim_keys[:-1] + (field.metadata("typeOfLevel"), "y", "x")
 
         if param not in metadata:
             metadata[param] = field.metadata(
@@ -89,25 +110,16 @@ def load_data(
     if not set(params) == data.keys():
         raise RuntimeError(f"Missing params: {set(params) - data.keys()}")
 
-    shapes = {}
+    result = {}
     for param, field_map in data.items():
-        steps, levels = zip(*field_map)
-        n_steps = len(set(steps))
-        n_levels = len(set(levels))
-        ny, nx = next(iter(field_map.values())).shape
-        shapes[param] = n_steps, n_levels, ny, nx
-
-    return {
-        param: xr.DataArray(
-            np.array([field_map.pop(key) for key in sorted(field_map)]).reshape(
-                shapes[param]
-            ),
-            coords=hcoords,
-            dims=["step", level_types[param], "y", "x"],
+        coords, shape = _gather_coords(field_map, dims[param])
+        result[param] = xr.DataArray(
+            np.array([field_map.pop(key) for key in sorted(field_map)]).reshape(shape),
+            coords=coords | hcoords,
+            dims=dims[param],
             attrs=metadata[param],
         )
-        for param, field_map in data.items()
-    }
+    return result
 
 
 def load_cosmo_data(
--- a/src/idpi/grib_decoder.py
+++ b/src/idpi/grib_decoder.py
@@ -1,26 +1,134 @@
 """Decoder for grib data."""
+# Standard library
+import os
+from contextlib import contextmanager
+from pathlib import Path
+
 # Third-party
-import cfgrib  # type: ignore
+import earthkit.data  # type: ignore
+import eccodes  # type: ignore
+import numpy as np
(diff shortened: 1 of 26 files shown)


src/idpi/grib_decoder.py
```diff
@@ -1,26 +1,131 @@
...
+            hcoords = {
+                dim: (("y", "x"), values) for dim, values in f.to_points().items()
+            }
+
+    if not set(params) == data.keys():
+        raise RuntimeError("Not all parameters were loaded")
+
+    return {
+        param: xr.DataArray(
+            np.array([levels.pop(k) for k in sorted(levels)]),
```
cfkanesan, line 92: The order by method provided by `earthkit.data` has a large performance impact for some reason. Thus, the ordering of the levels is done here.

src/idpi/grib_decoder.py
```diff
@@ -1,26 +1,134 @@
...
+        if param not in level_types:
+            level_types[param] = f.metadata("typeOfLevel")
+
+        if param not in metadata:
+            metadata[param] = f.metadata(
+                namespace=["ls", "geography", "parameter", "time"]
+            )
+
+        if hcoords is None and param == ref_param:
+            hcoords = {
```
cosunae, line 85: I think we can safely remove the horizontal coordinates. The design idea was to put all geographical information in a separate constant object, and attach here only information about staggering wrt the reference grid. 
I would start with nothing and later add it, for the time being we do not have a use case for this.

cosunae: On the contrary, I think we need information about the vertical coordinate. While all variables will be on the same horizontal grid, that is not the case for the vertical. 
https://hackmd.io/559zJIyMQSu_Qi2Sw81WRA
https://github.com/MeteoSwiss-APN/icon_data_processing_incubator/pull/56/files#diff-b8b3cc08fe5792c56c02926372a8e41bacbba127344a55635aedcb14d73f7d8eR47

cosunae (changes requested):

src/idpi/operators/pot_vortic.py
```diff
@@ -57,8 +57,13 @@ def fpotvortic(
     """
     # target coordinates
     deg2rad = np.pi / 180
-    lat = (rho_tot["latitude"] * deg2rad).astype(np.float32)
-    rlat = get_rotated_latitude(w)
+    lat = (rho_tot["lat"] * deg2rad).astype(np.float32)
+
+    geo = w.attrs["geography"]
+    ny = geo["Nj"]
+    lat_min = geo["latitudeOfFirstGridPointInDegrees"]
```
cosunae, line 64: this is an example of information that I think could come from a grid object that is passed to operators that need geographical information, that is constructed at setup.

src/idpi/grib_decoder.py
```diff
@@ -1,26 +1,134 @@
...
-        chunk_arg = {"chunks": {"generalVerticalLayer": chunk_size}}
 
-    # Note: dataset assignment is based on typeOfLevel in cfgrib
-    dss = cfgrib.open_datasets(
-        datafile,
-        backend_kwargs={"read_keys": ["typeOfLevel", "gridType"]},
-        encode_cf=("time", "geography", "vertical"),
-        **chunk_arg,
+@contextmanager
+def cosmo_grib_defs():
```
cosunae, line 15: this is a good change. I think this potentially might eliminate the need to use pytest markers: 
https://github.com/MeteoSwiss-APN/icon_data_processing_incubator/blob/main/pytest.ini

src/idpi/grib_decoder.py
cfkanesan: The latitude values are needed in the potential vorticity operator. Until we introduce the grid information object I thought that this was the most obvious place to put this information. In this PR, the vertical dimensions have a different name for different grid locations as they did before.

src/idpi/operators/pot_vortic.py
cfkanesan: Implementing that grid description object is not within the scope of this PR I believe.

src/idpi/grib_decoder.py
cosunae: I agree that we should do the grid info in a separate PR. But I was under the impression that the horizontal coordinates are not really used, since we seems to construct them directly from the grib rotated grid information: 
https://github.com/MeteoSwiss-APN/icon_data_processing_incubator/pull/59/files#diff-8bc8502e3870d03e9ad32dd60b42006509625d6e368d08306ed697c331946f46R66

Or where in pot vortic it is needed  ?

src/idpi/operators/pot_vortic.py
cosunae: yes, I agree

src/idpi/grib_decoder.py
cfkanesan: The geographic latitude is needed for the calculation of the Coriolis terms

tests/test_idpi/test_curl.py
```diff
@@ -20,25 +21,26 @@ def print_time(name):
...
+    dlat = geo["jDirectionIncrementInDegrees"]
+    nj = geo["Nj"]
+    lat_min = geo["latitudeOfFirstGridPointInDegrees"]
 
+    deg2rad = np.pi / 180
+    rlat = get_grid_coords(nj, lat_min, dlat, "y") * deg2rad
     total_diff = TotalDiff(dlon * deg2rad, dlat * deg2rad, ds["HHL"])
-    lat = ds["HHL"]["latitude"] * deg2rad
 
     with print_time("curl"):
```
cosunae, line 40: is this intentional or some debugging printouts?

cosunae (approved):

cosunae (changes requested):
the PR looks good to me, we would just need to resolve conflicts and I guess need to integrate the earthkit data work in the new tests of the main (flexpart)

src/idpi/operators/brn.py
```diff
@@ -15,8 +15,10 @@ def fbrn(p, t, qv, u, v, hhl, hsurf):
     nlevels = len(p.coords["generalVerticalLayer"])
 
     thetav = fthetav(p, t, qv)
-    thetav_sum = thetav.isel(generalVerticalLayer=slice(None, None, -1)).cumsum(
-        dim="generalVerticalLayer"
+    thetav_sum = (
```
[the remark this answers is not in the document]
cosunae: thanks, that was quite hidden (that the previous implementation was relying on it), indeed makes sense.

tests/test_idpi/test_curl.py
cfkanesan: It was originally intentional but I can remove it. Now the implementations are more or less equivalent since the introduction of the destagger ufuncs and diff operators.

cfkanesan:
You mean to replace the ifs data loader in the flexpart operator module ?

tests/test_idpi/test_curl.py
cosunae: I would remove it or rather have a concept for how to do logging and optionally print timings in all our operators. So that all operators have always the same behaviour, using for example a @operator decorator. But it is not part of this PR, we should create another issue for that

cosunae (approved):

PR #60, opened 2023-06-14
Misc fixes

cfkanesan:
- fix jenkins misreporting failures on the ifs marked tests
- package field mapping data file
- revert the removal of the index file due to the impact on the test time (>1h30 instead of 15 min)


pyproject.toml
```diff
@@ -34,6 +34,13 @@ documentation = "https://github.io/MeteoSwiss-APN/icon-data-processing-incubator
 # Format: <command> = "<package>.<module>:<function>"
 # icon-data-processing-incubator = "idpi.cli:main"
 
+[tool.setuptools.packages.find]
+namespaces = true
+where = ["src"]
+
+[tool.setuptools.package-data]
+"idpi.data" = ["*.yml"]
```
cfkanesan, line 42: https://setuptools.pypa.io/en/latest/userguide/datafiles.html#subdirectory-for-data-files

tests/test_idpi/test_flexpart.py
```diff
@@ -47,7 +46,7 @@ def test_flexpart():
     )
 
     loader = flx.ifs_data_loader(
-        (pathlib.Path(root_dir) / ".." / "share" / "field_mappings.yml").resolve()
+        files("idpi.data").joinpath("field_mappings.yml").open()
```
cfkanesan, line 49: https://setuptools.pypa.io/en/latest/userguide/datafiles.html#accessing-data-files-at-runtime

src/idpi/grib_decoder.py
```diff
@@ -11,7 +11,7 @@ def load_data(outds, fields, datafile, chunk_size=10):
     # Note: dataset assignment is based on typeOfLevel in cfgrib
     dss = cfgrib.open_datasets(
         datafile,
-        backend_kwargs={"read_keys": ["typeOfLevel", "gridType"], "indexpath": ""},
```
cosunae, line 14: this was introduced because there was a bug that happened when a indexpath was generated and used by different versions of eccodes. The default indexpath of cfgrib is hardcoded to the folder where data is, and can not be changed to a "local" cache of each run. The bug was not easy to debug which makes the use of the indexpath error prone

cfkanesan: I understand that this was introduced to prevent undefined behaviour but the impact on the build time makes it difficult to work with the rest of the repo. In any case, the earthkit data change will eliminate this code path soon. However, earthkit data is also writing some index file if I am not mistaken. We should check if it is subject to this issue.

cosunae (approved):

PR #62, opened 2023-06-26
Metadata

cfkanesan:
## Purpose

Introduce the proposed metadata model. 

## Code changes:

- Vertical coordinates are described through the `vcoord_type` attribute.
- Staggering is described through the `origin` attribute.
- The vertical dimension is named "z".
- Time coordinates are stored as datetime objects.

### DIFF
--- a/src/idpi/grib_decoder.py
+++ b/src/idpi/grib_decoder.py
@@ -11,7 +11,22 @@ import earthkit.data  # type: ignore
 import eccodes  # type: ignore
 import numpy as np
 import xarray as xr
-import yaml
+import yaml  # type: ignore
+
+
+DIM_MAP = {
+    "level": "z",
+    "perturbationNumber": "eps",
+    "step": "time",
+}
+VCOORD_TYPE = {
+    "hybrid": ("hybrid", 0.0),
+    "isobaricInPa": ("pressure", 0.0),
+    "generalVertical": ("model_level", -0.5),
+    "generalVerticalLayer": ("model_level", 0.0),
+    "surface": ("surface", 0.0),
+}
+_ifs_allowed = True
 
 
 @contextmanager
@@ -64,7 +79,7 @@ def _gather_tcoords(time_meta):
         if time is None:
             time = _parse_datetime(tm["dataDate"], tm["dataTime"])
 
-    return {"valid_time": ("step", valid_time), "time": time}
+    return {"valid_time": ("time", valid_time), "ref_time": time}
 
 
 def _extract_pv(pv):
@@ -72,9 +87,25 @@ def _extract_pv(pv):
         return {}
     i = len(pv) // 2
     return {
-        "ak": xr.DataArray(pv[:i], dims="hybrid"),
-        "bk": xr.DataArray(pv[i:], dims="hybrid"),
+        "ak": xr.DataArray(pv[:i], dims="z"),
+        "bk": xr.DataArray(pv[i:], dims="z"),
     }
+
+
+def _update_origin(metadata, ref_param):
+    x0_key = "longitudeOfFirstGridPointInDegrees"
+    y0_key = "latitudeOfFirstGridPointInDegrees"
+    ref_geo = metadata[ref_param]["geography"]
+    x0 = ref_geo[x0_key]
+    y0 = ref_geo[y0_key]
+    for m in metadata.values():
+        geo = m["geography"]
+        dx = geo["iDirectionIncrementInDegrees"]
+        dy = geo["jDirectionIncrementInDegrees"]
+        m["origin"] |= {
+            "x": np.round((geo[x0_key] - x0) / dx, 1),
+            "y": np.round((geo[y0_key] - y0) / dy, 1),
+        }
 
 
 def load_data(
@@ -112,7 +143,9 @@ def load_data(
     fs = earthkit.data.from_source("file", [str(p) for p in datafiles])
 
     if ref_param not in params:
-        raise ValueError(f"{ref_param} must be in {params}")
+        raise ValueError(f"{ref_param=} must be in {params=}")
+    if extract_pv is not None and extract_pv not in params:
+        raise ValueError(f"If set, {extract_pv=} must be in {params=}")
 
     hcoords = None
     pv = None
@@ -136,10 +169,16 @@ def load_data(
             time_meta.setdefault(param, {})[step] = field.metadata(namespace="time")
 
         if param not in dims:
-            dims[param] = dim_keys[:-1] + (field.metadata("typeOfLevel"), "y", "x")
+            dims[param] = tuple(DIM_MAP[d] for d in dim_keys) + ("y", "x")
 
         if param not in metadata:
             metadata[param] = field.metadata(namespace=["geography", "parameter"])
+            level_type = field.metadata("typeOfLevel")
+            vcoord_type, zshift = VCOORD_TYPE.get(level_type, (level_type, 0.0))
+            metadata[param] |= {
+                "vcoord_type": vcoord_type,
+                "origin": {"z": zshift},
+            }
 
         if hcoords is None and param == ref_param:
             hcoords = {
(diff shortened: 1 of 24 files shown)


src/idpi/grib_decoder.py
```diff
@@ -11,7 +11,22 @@
 import eccodes  # type: ignore
 import numpy as np
 import xarray as xr
-import yaml
+import yaml  # type: ignore
```
twicki, line 14: pip installing `pyyaml-include` gives you what you want here

src/idpi/grib_decoder.py
```diff
@@ -11,7 +11,22 @@
...
+    "step": "time",
+}
+VCOORD_TYPE = {
+    "generalVertical": ("model_level", -0.5),
+    "generalVerticalLayer": ("model_level", 0.0),
+    "hybrid": ("hybrid", 0.0),
+    "isobaricInPa": ("pressure", 0.0),
+    "surface": ("surface", 0.0),
+}
+_ifs_allowed = True
```
twicki, line 29: why does this need to be a global variable?

src/idpi/operators/destagger.py
```diff
@@ -84,42 +84,52 @@ def destagger(
...
     if dim == "x" or dim == "y":
-        return xr.apply_ufunc(
-            interpolate_midpoint,
-            field.reset_coords(drop=True),
-            input_core_dims=[[dim]],
-            output_core_dims=[[dim]],
-            kwargs={"extend": "left"},
-        ).transpose(*dims)
-    elif dim == "generalVertical":
+        if field.origin[dim] != 0.5:
```
twicki, line 105: we could extract the argument consistency checks from here to a separate call. That way it would be easier to extend and turn off if we want to

src/idpi/operators/pot_vortic.py
```diff
@@ -81,4 +81,6 @@ def fpotvortic(
         dt_dlam * curl1 + dt_dphi * (curl2 + cor2) - dt_dzeta * (curl3 + cor3)
     ) / rho_tot
 
+    out.attrs = theta.attrs
```
twicki, line 84: how was this not relevant before?

src/idpi/operators/support_operators.py
```diff
@@ -73,15 +73,12 @@ def init_field_with_vcoord(
     #       in the interface
     # attrs
     attrs = parent.attrs.copy()
-    attrs["GRIB_typeOfLevel"] = vcoord.type_of_level
-    if "GRIB_NV" in attrs:
```
twicki, line 77: why do we not need this update anymore?

src/idpi/operators/vertical_interpolation.py
```diff
@@ -253,36 +252,31 @@ def interpolate_k2theta(
     )
 
     # Check that typeOfLevel is supported and equal for field, th_field, and h_field
-    if supported_vc_type not in field.dims:
+    if field.vcoord_type != "model_level" or field.origin["z"] != 0.0:
```
twicki, line 255: what vcord type has a 0-origin and ist not model_level?

And if they are independent, what is the relevant test here?

tests/test_idpi/test_curl.py
```diff
@@ -37,12 +25,10 @@ def test_curl(data_dir):
     rlat = get_grid_coords(nj, lat_min, dlat, "y") * deg2rad
     total_diff = TotalDiff(dlon * deg2rad, dlat * deg2rad, ds["HHL"])
 
-    with print_time("curl"):
```
twicki, line 40: did we not acre about the times here anymore?

twicki (commented):
just a few understanding questions

src/idpi/grib_decoder.py
cfkanesan: It's a flag that prevents ifs data from being loaded after cosmo data since conflicting definitions are being cached by eccodes. However, now that I think about it there should and equivalent comso_allowed flag since the reverse situation is also possible.

src/idpi/operators/destagger.py
cfkanesan: The consistency checks definitely need to be extended at some point for all operators but this requires some additional thought regarding the design. That said, I don't understand the suggestion that you are making. Should it be a call that is done before the call to destagger? How would you ensure that it has been called from the point of view of this function?

src/idpi/operators/pot_vortic.py
cfkanesan: The handling of the attributes is not consistent throughout the library at the moment. The pot vortic output needs to be interpolated at theta values and integrated between pressure levels for the ninjo product so the vertical coordinate attributes are required.

src/idpi/operators/support_operators.py
cfkanesan: The `GRIB_` attributes were set by cfgrib and no longer exist. The `NV` attribute describes the number of hybrid levels and is only relevant if we were to write some field defined on hybrid levels to grib. In any case, it can be derived from the vertical coordinates I think.

src/idpi/operators/vertical_interpolation.py
cfkanesan: Here we want to raise if the field is not defined on the masspoints, i.e. model level and origin z = 0.

tests/test_idpi/test_curl.py
cfkanesan: @cosunae requested this to be removed in another PR

src/idpi/grib_decoder.py
twicki: two follow up questions here then:

Do we not have any way to clear the cache, in your warning you write to respawn the process. Is there no other control?
Now that this functionality requires state and access to constant dicts, would making it a class make sense?
2 can be an afterthought to this PR but we definitely need the flags going both directions

src/idpi/operators/destagger.py
twicki: I would do 
```
def destagger(...):
  check_field() -> checks on x,y and z based on the input, raise if none of those
  if x,y:
    apply_ufunc1
  elif z:
   appy_ufunc2
```

src/idpi/operators/pot_vortic.py
twicki: 👍

src/idpi/operators/support_operators.py
twicki: 👍

src/idpi/operators/vertical_interpolation.py
twicki: so if we want to check for definition on mass points, why is model_level a relevant check? isn't just the origin the relevant part here?

src/idpi/grib_decoder.py
cfkanesan: There's an API to reset the grib context object but the issue is that it triggers a segmentation fault when called.

> Now that this functionality requires state and access to constant dicts, would making it a class make sense?

The class would still need global/static variables to provide the kind of functionality that we are looking for. I think it would bring little benefit.

src/idpi/operators/destagger.py
cfkanesan: I see thanks for the snippet. The condition to check is that origin = 0.5 for x and y, and origin = -0.5 for z. The check would therefore have a similar structure to the current function but with the added indirection. I think this layout would make sense if there were more checks that are in common for all directions.

src/idpi/operators/vertical_interpolation.py
cfkanesan: The mass points are located on model levels. Fields that are defined on other vertical coordinates are therefore not located on the mass points.

cosunae: in principle, we only have staggered data on model levels, so vcoord_type ="ml". 
All data (I think) we generate (with Fieldextra) on theta or pressure levels is always colocated. 
However, after looking into the grib type of vertical coordinate systems: 
https://github.com/ecmwf/eccodes/blob/6efa9fc52899863d32311e5498f79abde8f303f3/definitions/grib2/typeOfLevelConcept.def#L34
We see that it is possible to have layer on theta and pressure levels. So probably it would make sense to keep the origin defined also for those vcoord_types (even if we do not use them at the moment) ?

twicki (approved):
LGTM

PR #66, opened 2023-08-04
Add time operator delta

cfkanesan:
## Purpose

Implement the time operator delta.

## Code changes:

- Add delta method in time_operators module.


tests/test_idpi/test_time_ops.py
```diff
@@ -0,0 +1,43 @@
...
+def test_delta(data_dir, fieldextra):
+    steps = np.arange(0, 16, 3)
+    dd, hh = np.divmod(steps, 24)
+    datafiles = [data_dir / f"lfff{d:02d}{h:02d}0000" for d, h in zip(dd, hh)]
+
+    ds = grib_decoder.load_cosmo_data(["TOT_PREC"], datafiles, ref_param="TOT_PREC")
+
+    tot_prec_03h = delta(ds["TOT_PREC"], np.timedelta64(3, "h"))
+    cond = np.logical_or(tot_prec_03h > 0.0, tot_prec_03h.isnull())
+    observed = tot_prec_03h.where(cond, 0.0)
```
petrabaumann, line 29: For the reader who is unfamiliar with fieldextra, one should maybe add a comment that fieldextra sets a missing value if the delta operation cannot be computed, and that in the reference product the result of the computation is explicitly clipped to avoid negative values due to numerical inaccuracy.

tests/test_idpi/fieldextra_templates/test_time_ops_delta.nl
```diff
@@ -0,0 +1,32 @@
...
+ earth_axis_small=6371229.
+/
+
+&Process
+in_file="{{ file.inputi }}"
+tstart=0, tstop=15, tincr=3, tlag=-3,0,3
+out_file="{{ file.output }}"
+out_type="NETCDF"
+/
+&Process in_field = "TOT_PREC", tag='tot_prec_03h', toper='delta,3,hour', poper="replace_cond,<0.,0." /
```
petrabaumann, line 30: The comment on clipping could also be added here.

src/idpi/operators/time_operators.py
```diff
@@ -19,3 +19,38 @@ def time_rate(var: xr.DataArray, dtime: np.timedelta64):
...
+    condition = coord - coord[0] == dtime
+
+    try:
+        [index] = np.nonzero(condition.values)
+    except ValueError:
+        msg = "Provided dtime is not a multiple of the time step."
+        raise ValueError(msg)
+
+    result = field - field.shift(time=index.item())
+    result.attrs = field.attrs
```
petrabaumann, line 55: If we want to handle meta data strictly, we should perhaps introduce time bounds, and set the cell_methods attribute with respect to time (or code GRIB2-like attributes like typeOfStatisticalProcessing and lengthOfTimeRange).

petrabaumann (commented):
Fine for me - I only have questions on how far we should go to re-set meta information on time aggregation here.

tests/test_idpi/fieldextra_templates/test_time_ops_delta.nl
cfkanesan: Should I use the Jinja comment syntax or is there one that is supported by fieldextra?

tests/test_idpi/test_time_ops.py
cfkanesan: Why are negative values assumed to be result of numerical inaccuracy? Is it due to the nature of the field, i.e. precipation can not be negative?

src/idpi/operators/time_operators.py
cfkanesan: This sounds like a good idea but I don't know how this should be specified. In theory, we are not bound by grib2 semantics but it would be smart not to drift too far away.

petrabaumann: We should keep this issue in mind for further dicussion of the meta data model.

petrabaumann: I saw that you made a corresponding comment on the Confluence page dedicated to the meta data model.

tests/test_idpi/test_time_ops.py
petrabaumann: PS: Negative values in time differences of precipitation may also occur due to the GRIB compression.

petrabaumann (approved):

cfkanesan:
@petrabaumann Thanks for the clarifications!

PR #68, opened 2023-08-08
Time ops minmaxavg

cfkanesan:
## Purpose

Implement additional time operators.

## Code changes:

- Add `nsteps` and `resample` functions to the `time_operators` module.

### DIFF
--- a/src/idpi/operators/time_operators.py
+++ b/src/idpi/operators/time_operators.py
@@ -21,9 +21,15 @@ def time_rate(var: xr.DataArray, dtime: np.timedelta64):
     return result
 
 
-def _nsteps(valid_time, dtime):
-    condition = valid_time - valid_time[0] == dtime
+def _nsteps(valid_time: xr.DataArray, dtime: np.timedelta64) -> int:
+    dt = valid_time.diff(dim="time")
+    uniform = np.all(dt == dt[0]).item()
 
+    if not uniform:
+        msg = "Given field has an irregular time step."
+        raise ValueError(msg)
+
+    condition = valid_time - valid_time[0] == dtime
     try:
         [index] = np.nonzero(condition.values)
     except ValueError:
--- a/src/idpi/operators/time_operators.py
+++ b/src/idpi/operators/time_operators.py
@@ -21,6 +21,18 @@ def time_rate(var: xr.DataArray, dtime: np.timedelta64):
     return result
 
 
+def _nsteps(valid_time, dtime):
+    condition = valid_time - valid_time[0] == dtime
+
+    try:
+        [index] = np.nonzero(condition.values)
+    except ValueError:
+        msg = "Provided dtime is not a multiple of the time step."
+        raise ValueError(msg)
+
+    return index.item()
+
+
 def delta(field: xr.DataArray, dtime: np.timedelta64) -> xr.DataArray:
     """Compute difference for a given delta in time.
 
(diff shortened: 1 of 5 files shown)


src/idpi/operators/time_operators.py
```diff
@@ -34,23 +52,149 @@ def delta(field: xr.DataArray, dtime: np.timedelta64) -> xr.DataArray:
...
+    -------
+    xr.DataArray
+        The field maximum for the given time delta.
+
+    """
+    nsteps = _nsteps(field.valid_time, dtime)
+    return field.rolling(time=nsteps).max()
+
+
+def avg(field: xr.DataArray, dtime: np.timedelta64) -> xr.DataArray:
```
cosunae, line 122: In order to distinguish this operator from the regular avg, min, max, and having to overload, I was thinking that moving average, or rolling average is a more precise function name. Also dtime -> dtime_window would help understand the behaviour. 

For example: 
```
rolling_avg(field: xr.Darray, dtime_window: np.timedelta64)
```

src/idpi/operators/time_operators.py
```diff
@@ -34,23 +52,149 @@ def delta(field: xr.DataArray, dtime: np.timedelta64) -> xr.DataArray:
...
+    """Compute average aggregate for a given delta in time.
```
cosunae, line 123:
```suggestion
    """Compute average aggregate for a given time window.
```

src/idpi/operators/time_operators.py
```diff
@@ -34,23 +52,149 @@ def delta(field: xr.DataArray, dtime: np.timedelta64) -> xr.DataArray:
...
+    -------
+    xr.DataArray
+        The field sum for the given time delta.
+
+    """
+    nsteps = _nsteps(field.valid_time, dtime)
+    return field.rolling(time=nsteps).sum()
+
+
+def resample(field: xr.DataArray, period: np.timedelta64) -> xr.DataArray:
```
cosunae, line 174: I suggest dt instead of period. More generically (not only for time) would be sample interval

src/idpi/operators/time_operators.py
cosunae: alternatively, I wonder the value of these functions, since I think 
```
field.rolling(time=nsteps).max()
```
is the most readable version and does not require writing multiple function for the various operators min, max, etc. 

Therefore I would suggest an alternative, making nsteps a public function, which can be used directly in the products before the rolling function.

tests/test_idpi/test_time_ops.py
```diff
@@ -43,3 +43,32 @@ def test_delta(data_dir, fieldextra):
     expected = xr.concat([fx_ds["tot_prec_03h"] for fx_ds in fx_ds_h], dim="time")
 
     assert_allclose(observed, expected.transpose("epsd_1", "time", ...))
+
+
+def test_max(data_dir, fieldextra):
```
cosunae, line 48: It would be nice some of the raises, by constracting some artificial arrays with non uniform time coordinates or non multiple steps.

src/idpi/operators/time_operators.py
```diff
@@ -21,6 +21,24 @@ def time_rate(var: xr.DataArray, dtime: np.timedelta64):
...
+    dt = valid_time.diff(dim="time")
+    uniform = np.all(dt == dt[0]).item()
+
+    if not uniform:
+        msg = "Given field has an irregular time step."
+        raise ValueError(msg)
+
+    condition = valid_time - valid_time[0] == dtime
+    try:
+        [index] = np.nonzero(condition.values)
```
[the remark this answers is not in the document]
cfkanesan: The tuple unpacking raises when the number of elements does not match.

src/idpi/operators/time_operators.py
cfkanesan: I think interval is more explicit than dt

cosunae (approved):

PR #69, opened 2023-08-15
Add lateral operators fill_undef and disk_avg

cfkanesan:
## Purpose

Add lateral operators fill_undef and disk_avg.

## Code changes:

- Add module `lateral_operators` containing the `fill_undef` and `disk_avg` functions.

### DIFF
--- a/src/idpi/operators/lateral_operators.py
+++ b/src/idpi/operators/lateral_operators.py
@@ -1,3 +1,5 @@
+"""Lateral operators."""
+
 # Standard library
 from typing import Literal
 
@@ -11,6 +13,38 @@ def compute_weights(
     window_type: Literal["exp", "const"],
     window_shape: Literal["disk", "square"],
 ) -> xr.DataArray:
+    """Compute weights for a convolution kernel.
+
+    Parameters
+    ----------
+    window_size : int
+        Number of grid cells the kernel. Must be an odd number.
+    window_type : {"exp", "const"}
+        The window type.
+
+        Supported types:
+         - exp: Exponential drop from the center of the window.
+         - const: All weights are set to 1.
+
+    window_shape : {"disk", "square"}
+        The window shape.
+
+        Supported shapes:
+         - disk: The kernel is confined to the circle inscribing the window.
+         - square: The kernel is defined on the entire window.
+
+    Raises
+    ------
+    ValueError
+        if `window_size` is not an odd number.
+
+    Returns
+    -------
+    xarray.DataArray
+        Weights corresponding to the `window_type` within the given `window_shape`
+        and zero outside of the shape.
+
+    """
     n = window_size
 
     if n % 2 != 1:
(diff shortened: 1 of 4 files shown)


src/idpi/operators/lateral_operators.py
```diff
@@ -0,0 +1,192 @@
...
+
+    undef = windows.isnull()
+    frac_undef = undef.where(mask).where(in_bnds_x).where(in_bnds_y).mean(weights.dims)
+    return frac_undef <= frac_val
+
+
+def fill_undef(field: xr.DataArray, radius: int, frac_val: float) -> xr.DataArray:
+    """Fill undefined values.
+
+    Replace undefined values with a value derived from neighbourhood.
```
cosunae, line 116: can you specify in the doc how the value from the neighbourhood is computed?

src/idpi/operators/lateral_operators.py
```diff
@@ -0,0 +1,192 @@
...
+
+    Parameters
+    ----------
+    field : xarray.DataArray
+        Input field.
+    radius : int
+        Radius of the convolution window.
+    frac_val : float
+        Threshold of acceptable ratio of undefined values in window.
```
cosunae, line 125: it is unclear from the doc what happens if this threshold is surpassed

src/idpi/operators/lateral_operators.py
```diff
@@ -0,0 +1,192 @@
...
+
+# Standard library
+from typing import Literal
+
+# Third-party
+import numpy as np
+import xarray as xr
+
+
+def compute_weights(
```
cosunae, line 11: perhaps we could unittest this function ? for some of the use cases and compare against a numpy array of the expected values

cosunae: I mean an isolated test, out of the regression that is using fieldextra

src/idpi/operators/lateral_operators.py
```diff
@@ -0,0 +1,192 @@
...
+    Returns
+    -------
+    xarray.DataArray
+        Boolean mask indicating where the undefined value ratio in the convolution
+        kernel is within the acceptable threshold.
+
+    """
+    mask = weights > 0
+
+    loc_x = windows.x + windows.win_x - windows.sizes["win_x"] // 2
```
cosunae, line 102: this block is hard to follow, not sure if some comments would help to digest it better?

cosunae:
I generally quite like it, I did run it to watch some of the weight patterns for the convolution, and seems all good to me. 
I just have some comments to improve mostly the documentation

src/idpi/operators/lateral_operators.py
cfkanesan: For each window, `x` is the coordinate of the center of the window in the parent field. `win_x` is the local coordinate of the current element in the window. `loc_x` represents the location of the window element in the coordinates of the parent field. This is needed to filter out nans that are introducing by the padding of the field by the rolling method. Only the nans that are present within the bound of the field should count towards the `frac_val` threshold.

cfkanesan: Should I add this paragraph as a comment?

src/idpi/operators/lateral_operators.py
cfkanesan: The value for that window remains undefined.

src/idpi/operators/lateral_operators.py
cfkanesan: It's essentially a Gaussian filter with sigma = 1.

src/idpi/operators/lateral_operators.py
cosunae: yes, I would add it to the comment

cfkanesan: I have added this in the latest commit

cosunae: sorry, I didnt realize

cosunae (approved):

PR #70, opened 2023-08-16
Add hzerocl extrapolate option

cfkanesan:
## Purpose

Add the extrapolate below ground option to the zero degree C isotherm height computation.

## Code changes:

- Add flag `extrapolate` to `fhzerocl` operator. When set to True, the algorithm extrapolates the temperature values in the last 2 model layers down to the zero degree intercept.

### DIFF
--- a/src/idpi/operators/hzerocl.py
+++ b/src/idpi/operators/hzerocl.py
@@ -2,26 +2,37 @@
 
 This is done without extrapolation below model orography.
 """
+# Standard library
+from typing import cast
+
 # Third-party
+import numpy as np
 import xarray as xr
 
 # First-party
 from idpi.operators.destagger import destagger
 
 
-def fhzerocl(t: xr.DataArray, hhl: xr.DataArray) -> xr.DataArray:
+def fhzerocl(
+    t: xr.DataArray, hhl: xr.DataArray, extrapolate: bool = False
+) -> xr.DataArray:
     """Height of the zero deg C isotherm in m amsl.
 
     The algorithm searches from the top of the atmosphere downwards.
-    No extrapolation below the earth's surface is done.
 
-    Args:
-        t (xarray.DataArray): air temperature in K
-        hhl (xarray.DataArray): heights of the interfaces between
-            vertical layers in m amsl
+    Parameters
+    ----------
+    t : xr.DataArray
+        Air temperature in K.
+    hhl : xr.DataArray
+        Heights of the interfaces between vertical layers in m amsl.
+    extrapolate : bool
+        Allow the extrapolation of the search below the lowest model layer.
 
-    Returns:
-        xarray.DataArray: height of the zero deg C isotherm in m amsl
+    Returns
+    -------
+    xr.DataArray
+        Height of the zero deg C isotherm in m amsl.
 
     """
     # Physical constants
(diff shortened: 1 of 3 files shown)


src/idpi/operators/hzerocl.py
```diff
@@ -34,13 +44,23 @@ def fhzerocl(t: xr.DataArray, hhl: xr.DataArray) -> xr.DataArray:
...
+    maxind = cast(xr.DataArray, height0.fillna(-1).argmax(dim="z"))
+
+    if extrapolate:
+        # The full column is below freezing
+        below_ground = (t < t0).all("z")
+        # Temperature is increasing with decreasing altitude
+        positive_dt = t[{"z": -1}] - t[{"z": -2}] > 1e-10
+        # Allow t0 to be outside of the [height1, height2] range
+        cond = np.logical_and(below_ground, positive_dt)
+        maxind = maxind.where(~cond, -1)
```
victoria-cherkas, line 60: Does this mean that maxind = -1 when entire column is below freezing and temp is increasing with decreasing altitude? New to this syntax.

src/idpi/operators/hzerocl.py
```diff
@@ -1,27 +1,37 @@
...
-        hhl (xarray.DataArray): heights of the interfaces between
-            vertical layers in m amsl
+    Parameters
+    ----------
+    t : xr.DataArray
+        Air temperature in K.
+    hhl : xr.DataArray
+        Heights of the interfaces between vertical layers in m amsl.
+    extrapolate : bool, optional
+        Allow the extrapolation of the search below the lowest model layer.
```
victoria-cherkas, line 28: Perhaps this description of extrapolate can be improved because still not entirely sure why it is needed and what it achieves. Do we have the temperatures below ground in `t`?

src/idpi/operators/hzerocl.py
```diff
@@ -34,13 +44,23 @@ def fhzerocl(t: xr.DataArray, hhl: xr.DataArray) -> xr.DataArray:
...
     # The previous condition can be satisfied on multiple levels.
     # Take the k indices of the maximum height value where the condition is satisfied
-    maxind: xr.DataArray = height2.fillna(-1).argmax(dim="z")  # type: ignore
+    maxind = cast(xr.DataArray, height0.fillna(-1).argmax(dim="z"))
+
+    if extrapolate:
+        # The full column is below freezing
+        below_ground = (t < t0).all("z")
+        # Temperature is increasing with decreasing altitude
+        positive_dt = t[{"z": -1}] - t[{"z": -2}] > 1e-10
```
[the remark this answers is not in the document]
cfkanesan: They are the ones that are closer to the surface

src/idpi/operators/hzerocl.py
cfkanesan: No the temperature field does not go below the ground. Perhaps there are some soil models that keep track of the underground temperature but those not considered here to my knowledge. Fieldextra proposes an option to extend the gradient of the last 2 model layers below to search for the zero degree C intercept. It's unclear to me whether this information is accurate but it's something that fieldextra does and we need it to reproduce the results of the product.

victoria-cherkas: If the lowest model layers reach the surface, is 'below' not then underground? Or do the lowest model layers not necessarily go to the surface? Perhaps my understanding is not 100% necessary, but think the descriptions should be very clear.

cfkanesan: The lowest model layers do not necessarily reach the surface due to the way that the orography is set up but they may also intersect it. Therefore, the notion of underground depends a bit on the reference. However, the zero degree C intercept may be several kilometers below the lowest model layer and that's definitely underground.

victoria-cherkas: Ok interesting, not how I thought the model layers worked.

victoria-cherkas (approved):

PR #72, opened 2023-08-26
Grib reader

cosunae:
## Purpose

Refactor the grib reader so that parameters can be loaded in individual functions that can later be dask delayed. 

## Code changes:
- implements the grib reader with a class instead of standalone functions
- introduce the option to dask delay the load functions
- reference field is loaded first, before any other parameter. 
- one record found of the reference parameter is enough to derive the grid reference information, therefore the full field is not constructed. 
- implement a Grid class to store the grid reference attributes

## Performance impact
these changes introduce a performance deterioration. The jenkins tests increased timing from ~3.30m to ~6.30m. 
This is due to the fact that with earthkit-data it is more efficient to load a collection of data fields at once, i.e. traversing only once the sequence of grib records contained in a grib file. That means, random access to fields is less efficient. And this PR split the load of a data field collection into individual tasks each loading one field, requiring then random access to data field loads. 
However, the final architecture for performance should be measured against a system that can efficiently provide random access  to data fields, i.e. FDB. Additionally, we should measure the effect on an entire product collection parallelize. Since often when parallelizing an algorithm (to make it thread safe or parallel) it incurs into a performance hit on the sequential code performance. 
I quickly measured the impact using the indexing of earhkit-data (#75 & #76). But I observed no difference

## Additional notes to reviewer

This PR belongs to a series to implement product structure based on dask delay. The order of dependency of these PRs follow: 

| idpi  |  idpi_products |
|------ | ----------------|
|grib_reader (#72)   ||
|first_product (#73)  | <----   first_product (https://github.com/MeteoSwiss-APN/idpi-products/pull/6)|
|                                |         support_various_datasets (https://github.com/MeteoSwiss-APN/idpi-products/pull/7)|
|relhum         (#74)   | <----   add_cosmo2e_products (https://github.com/MeteoSwiss-APN/idpi-products/pull/8)|
                                   

## Checklist

- [x] New operators are properly tested

## Review

### DIFF
--- a/requirements/requirements.yaml
+++ b/requirements/requirements.yaml
@@ -42,7 +42,7 @@ dependencies:
   - types-PyYAML>=6.0.10
   - pip:
     # development
-    - copier>=7.0
+    - copier>=7.0,<8
     - flake8-pyproject>=1.2
     - sphinx-mdinclude>=0.5
     - pydantic<2.0.0
--- a/requirements/requirements.yaml
+++ b/requirements/requirements.yaml
@@ -45,4 +45,3 @@ dependencies:
     - copier>=7.0,<8
     - flake8-pyproject>=1.2
     - sphinx-mdinclude>=0.5
-    - pydantic<2.0.0
--- a/requirements/requirements.yaml
+++ b/requirements/requirements.yaml
@@ -40,9 +40,9 @@ dependencies:
   - types-toml>=0.10
   - yaml>=0.2.1
   - types-PyYAML>=6.0.10
-  - pydantic<2.0.0
   - pip:
     # development
     - copier>=7.0
     - flake8-pyproject>=1.2
     - sphinx-mdinclude>=0.5
+    - pydantic<2.0.0
--- a/requirements/requirements.yaml
+++ b/requirements/requirements.yaml
@@ -40,6 +40,7 @@ dependencies:
   - types-toml>=0.10
   - yaml>=0.2.1
   - types-PyYAML>=6.0.10
+  - pydantic<2.0.0
   - pip:
     # development
     - copier>=7.0
--- a/src/idpi/grib_decoder.py
+++ b/src/idpi/grib_decoder.py
@@ -98,7 +98,7 @@ def _extract_pv(pv):
 class Grid:
     """Coordinates of the reference grid.
 
-    Parameters
+    Attributes
     ----------
     lon: xr.DataArray
         2d array with longitude of geographical coordinates
--- a/src/idpi/grib_decoder.py
+++ b/src/idpi/grib_decoder.py
@@ -176,25 +176,21 @@ class GribReader:
         if field is None:
             msg = f"reference field, {ref_param=} not found in {self._datafiles=}"
             raise RuntimeError(msg)
-            lonlat_dict = {
-                geo_dim: xr.DataArray(dims=("y", "x"), data=values)
-                for geo_dim, values in field.to_latlon().items()
-            }
+        lonlat_dict = {
+            geo_dim: xr.DataArray(dims=("y", "x"), data=values)
+            for geo_dim, values in field.to_latlon().items()
+        }
 
-            grid = Grid(
-                lonlat_dict["lon"],
-                lonlat_dict["lat"],
-                *field.metadata(
-                    "longitudeOfFirstGridPointInDegrees",
-                    "latitudeOfFirstGridPointInDegrees",
-                ),
-            )
-
-            return grid
-
-        raise RuntimeError(
-            f"reference field, {ref_param=} not found in {self._datafiles=}"
+        grid = Grid(
+            lonlat_dict["lon"],
+            lonlat_dict["lat"],
+            *field.metadata(
+                "longitudeOfFirstGridPointInDegrees",
+                "latitudeOfFirstGridPointInDegrees",
+            ),
         )
+
+        return grid
 
     def _load_pv(self, pv_param: str):
         if not self._ifs:
--- a/src/idpi/grib_decoder.py
+++ b/src/idpi/grib_decoder.py
@@ -1,7 +1,7 @@
 """Decoder for grib data."""
 # Standard library
 import datetime as dt
-import os
+import sys
 from contextlib import contextmanager
 from importlib.resources import files
 from pathlib import Path
@@ -32,8 +32,7 @@ _cosmo_allowed = True
 @contextmanager
 def cosmo_grib_defs():
     """Enable COSMO GRIB definitions."""
-    prefix = os.environ["CONDA_PREFIX"]
-    root_dir = Path(prefix) / "share"
+    root_dir = Path(sys.prefix) / "share"
     paths = (
         root_dir / "eccodes-cosmo-resources/definitions",
         root_dir / "eccodes/definitions",
--- a/src/idpi/grib_decoder.py
+++ b/src/idpi/grib_decoder.py
@@ -171,7 +171,6 @@ class GribReader:
             ref_param = mapping[ref_param]["ifs"]["name"]
 
         fs = earthkit.data.from_source("file", [str(p) for p in self._datafiles])
-
         it = iter(fs.sel(param=ref_param))
         field = next(it, None)
         if field is None:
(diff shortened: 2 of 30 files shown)


src/idpi/grib_decoder.py
```diff
@@ -91,227 +94,297 @@ def _extract_pv(pv):
...
+
+        """
+        if self._ifs:
+            mapping_path = files("idpi.data").joinpath("field_mappings.yml")
+            mapping = yaml.safe_load(mapping_path.open())
+            ref_param = mapping[ref_param]["ifs"]["name"]
+
+        fs = earthkit.data.from_source("file", [str(p) for p in self._datafiles])
+
+        for field in fs.sel(param=ref_param):
```
cfkanesan, line 171:
```suggestion
        it = iter(fs.sel(param=ref_param))
        field = next(it, None)
        if field is None:
            msg = f"reference field, {ref_param=} not found in {self._datafiles=}"
            raise RuntimeError(msg)
```

src/idpi/grib_decoder.py
```diff
@@ -91,227 +94,297 @@ def _extract_pv(pv):
...
+                        "z": zshift,
+                        "x": np.round(
+                            (geo["longitudeOfFirstGridPointInDegrees"] % 360 - x0) / dx,
+                            1,
+                        ),
+                        "y": np.round(
+                            (geo["latitudeOfFirstGridPointInDegrees"] - y0) / dy, 1
+                        ),
+                    },
+                }
```
cfkanesan, line 255: I think this was better in a function. Now the indent level is a bit excessive

src/idpi/grib_decoder.py
```diff
@@ -91,227 +94,297 @@ def _extract_pv(pv):
...
+
+        if not field_map:
+            raise RuntimeError(f"requested {param=} not found.")
+
+        coords, shape = _gather_coords(field_map, dims)
+        tcoords = _gather_tcoords(time_meta)
+        hcoords = {
+            "lon": (("y", "x"), self._grid.lon.data),
+            "lat": (("y", "x"), self._grid.lat.data),
```
cfkanesan, line 264: Why not use the DataArray objects?

tests/test_idpi/test_grib_decoder.py
```diff
@@ -0,0 +1,27 @@
...
+from unittest.mock import Mock
+from unittest.mock import patch
+
+# Third-party
+import pytest
+
+# First-party
+from idpi.grib_decoder import GribReader
+
+earthkit = Mock()
```
cfkanesan, line 11: Is this necessary?

tests/test_idpi/test_grib_decoder.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+@patch("idpi.grib_decoder.earthkit.data.from_source")
+def test_ref_param_not_found(mock_from_source):
```
cfkanesan, line 15: I don't really get the purpose of this test. The grib field iterator is replaced by a mock object which has a very different behaviour from the real object.

cfkanesan (commented):
I think the changes are pretty good, I left some comments throughout the diff. One thing to mention is the performance impact notably that the test run time has doubled. Hopefully this will be resolved by using FDB as a source but it underlines the need to have a tracking of the performance.

src/idpi/grib_decoder.py
```diff
@@ -91,227 +94,297 @@ def _extract_pv(pv):
...
+        ------
+        ValueError
+            if the grid can not be constructed from the ref_param
+
+        """
+        self._datafiles = datafiles
+        self._ifs = ifs
+        self._delayed = dask.delayed if delay else (lambda x: x)
+        if not self._ifs:
+            with cosmo_grib_defs():
```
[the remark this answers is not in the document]
cosunae: oh, yes, I forgot that global variable. done

src/idpi/grib_decoder.py
```diff
@@ -91,227 +94,297 @@ def _extract_pv(pv):
...
-        m["origin"] |= {
-            "x": np.round((geo[x0_key] % 360 - x0) / dx, 1),
-            "y": np.round((geo[y0_key] - y0) / dy, 1),
-        }
+@dc.dataclass
+class Grid:
+    lon: xr.DataArray
+    lat: xr.DataArray
+    longitudeOfFirstGridPointInDegrees: float
+    latitudeOfFirstGridPointInDegrees: float
```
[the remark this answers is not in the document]
cosunae: done

tests/test_idpi/test_grib_decoder.py
cosunae: no, good catch. It is a leftover from a previous attempt

src/idpi/grib_decoder.py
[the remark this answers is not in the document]
cfkanesan: maybe this mechanism needs to be improved because it's quite easy to forget it as it is implemented now. This would be the subject for a new PR however.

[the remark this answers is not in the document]
cosunae: agreed

tests/test_idpi/test_grib_decoder.py
cosunae: yes, you are right, a None would not be iterable. 
I added the explicit return_value = []. Is it more correct now? 

I agree with you that the test does not bring much value. 
We had already a task to reduce the domain size for testing of operators. 
If we create a tiny dataset 2x2 grid points, for example. We could easily test in a relatively quick way this on real data. We could also test construction of metadata and other functions. 
I can create a issue for this, I guess it would be for another PR. 
The question is should I then remove this tests?

src/idpi/grib_decoder.py
```diff
@@ -91,227 +94,303 @@ def _extract_pv(pv):
...
+        if not self._ifs:
+            raise ValueError("load_pv only available for IFS data")
+        fs = earthkit.data.from_source("file", [str(p) for p in self._datafiles]).sel(
+            param=pv_param
+        )
+
+        for field in fs:
+            return field.metadata("pv")
+
+    def _construct_metadata(self, field: typing.Any):
```
cosunae, line 204: I tried here to set the type to earthkit.data.xxx.GribField, however that requires loading the corresponding module of earthkit which causes that our current mechanism of setting the eccodes definitions path to fail. One can see that in the previous builds of the jenkins plans for this PR.

tests/test_idpi/test_grib_decoder.py
cfkanesan: With the return value it should be fine. I usually spec the mocks to make sure that they don't allow undefined attribute access but that's overkill in this case.

I think you can leave the test but add a note in the ticket to remove the patch once the cropped data is available.

src/idpi/grib_decoder.py
```diff
@@ -91,227 +94,303 @@ def _extract_pv(pv):
...
-    x0 = ref_geo[x0_key] % 360
-    y0 = ref_geo[y0_key]
-    for m in metadata.values():
-        geo = m["geography"]
+@dc.dataclass
+class Grid:
+    lon: xr.DataArray
+    lat: xr.DataArray
+    lon_first_grid_point: float
+    lat_first_grid_point: float
```
cfkanesan, line 102: Maybe there should be some hint that the 2 first attributes are in the geographical CRS and the others are in the rotated CRS.

cosunae: done, I add the doc in the style of a method, not sure if we have some other pattern to doc dataclasses

src/idpi/grib_decoder.py
```diff
@@ -91,227 +94,303 @@ def _extract_pv(pv):
...
+        Grid
+            reference grid
+
+        """
+        if self._ifs:
+            mapping_path = files("idpi.data").joinpath("field_mappings.yml")
+            mapping = yaml.safe_load(mapping_path.open())
+            ref_param = mapping[ref_param]["ifs"]["name"]
+
+        fs = earthkit.data.from_source("file", [str(p) for p in self._datafiles])
```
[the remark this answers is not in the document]
cosunae: indeed

tests/test_idpi/test_grib_decoder.py
cosunae: done, I opened a ticket https://meteoswiss.atlassian.net/browse/APNRZ-200

src/idpi/grib_decoder.py
cfkanesan: https://numpydoc.readthedocs.io/en/latest/format.html#documenting-classes

cfkanesan (approved):

requirements/requirements.yaml
```diff
@@ -45,3 +45,4 @@ dependencies:
     - copier>=7.0
     - flake8-pyproject>=1.2
     - sphinx-mdinclude>=0.5
+    - pydantic<2.0.0
```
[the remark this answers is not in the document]
cfkanesan: Actually this change is no longer needed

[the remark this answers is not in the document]
cosunae: yes, removed

PR #73, opened 2023-08-26
First product

cosunae:
## Purpose

Introduce a base class for products. The purpose in this PR is to provide through that base class infrastructure to automatically dask delay product runs. In the future it can also be used for other common infrastructure for products like logging, or perf monitoring (measure run of product)

Products should then follow a contract with the base class, like they should specify input_fields attribute, that list the input fields that need to be loaded. Today that is a collection of names, but in the future the data requests specification will probably require more complex requests (like mars requests).

## Code changes:

- implementation of product base class
- load function in the grib_decoder that accepts a collection of products (from where the input data collection is extracted).
 
## Additional notes to reviewer

This PR belongs to a series to implement product structure based on dask delay. The order of dependency of these PRs follow: 

| idpi  |  idpi_products |
|------ | ----------------|
|grib_reader (#72)   ||
|first_product (#73)  | <----   first_product (https://github.com/MeteoSwiss-APN/idpi-products/pull/6)|
|                                |         support_various_datasets (https://github.com/MeteoSwiss-APN/idpi-products/pull/7)|
|relhum         (#74)   | <----   add_cosmo2e_products (https://github.com/MeteoSwiss-APN/idpi-products/pull/8)|

## Checklist

- [x] New operators are properly tested

Additionally, if the PR updates the version of the package

- [ ] The new version is properly set
- [ ] A Tag will be created after the bump

## Review


src/idpi/product.py
```diff
@@ -0,0 +1,58 @@
...
+    """Register methods that are (dask) delayed for caching."""
+
+    def __init__(self, delay: bool = False):
+        self.regdict: dict[str, typing.Any] = {}
+        self._delayed = dask.delayed if delay else lambda x: x
+
+    def reg(self, fn, *arg):
+        key = list(
+            accumulate([hash(id(x)) for x in arg], lambda x, acc: hash(x ^ acc))
+        )[-1]
```
cfkanesan, line 23: Consider using a frozenset of the ids and hashing that. Note that the pattern `list(accumulate(...))[-1]` is equivalent to `reduce(...)`.

src/idpi/product.py
```diff
@@ -0,0 +1,58 @@
...
+        if not reg:
+            self.reg = Register()
+        else:
+            self.reg = reg
+        self._base_delayed = dask.delayed if delay_entire_product else lambda x: x
+
+    # avoid a possible override from inheriting classes
+    @property
+    def delay_entire_product(self):
+        return self._base_delayed
```
cfkanesan, line 47: Can you explain why this is needed?

src/idpi/product.py
cosunae: great suggestion thanks

src/idpi/product.py
cosunae: The main idea is that the base class will control the delay of entire products run, i.e. the situation where the entire product processing goes into one single task. I think this should cover most of the cases. 
However sometimes, we might want to have more fine grained control of the tasking, in case various products are sharing some intermediate computations. 
In that case, we would like to not introduce the delay at the product level (that is this property should be set to False), and the child class has the responsibility of delaying the various functions and operations in a more fine grained controled manner. 
The child class has another delay parameter which allows to activate/deactivate delaying tasks. So that we can also run products without dask delaying any computation. 
https://github.com/MeteoSwiss-APN/idpi-products/pull/6/files#diff-832bf11faf4adb36354b27b36199692f18199af6252edff013a54c9db72fec79R17

So the main intention here was to make this delay_entire_product a property, to avoid that the child classes overwrite the value of it, since there are similar attributes in childern. So that this can only be set at the init. 
Does it answer the question, does it make sense?

src/idpi/product.py
```diff
@@ -0,0 +1,56 @@
...
+    """Registry for operators that can be used to cache operations in dask."""
+
+    def __init__(self, delay: bool = False):
+        self.regdict: dict[str, typing.Any] = {}
+        self._delayed = dask.delayed if delay else lambda x: x
+
+    def reg(self, fn, *arg):
+        key = frozenset([id(x) for x in arg])
+
+        return self.regdict.setdefault(key, self._delayed(fn)(*arg))
```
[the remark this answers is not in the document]
cosunae: I did a test with the following code: 
```
    in_fields = ["P", "T", "QV", "U", "V", "HHL", "HSURF"]
    ds = reader.load_cosmo_data(in_fields)

    ds2 = reader.load_cosmo_data(in_fields)

    for field in in_fields:
        ds[field] = ds[field] + ds2[field]
```

and indeed I get the desired behaviour by passing pure=True to the dask.delayed function, so I will change this and remove this register class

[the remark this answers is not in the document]
cosunae: brn_graph
brn_graph_pure

src/idpi/product.py
cfkanesan: Following our discussion, it seems like the `delay_entire_product` property is not intended to be used other than to indicate that the `_base_delay` should not be overridden in the child classes

src/idpi/product.py
```diff
@@ -0,0 +1,40 @@
...
+from abc import abstractmethod
+from functools import partial
+
+# Third-party
+import dask
+
+
+@dc.dataclass
+class ProductDescriptor:
+    input_fields: list[str]
```
cfkanesan, line 15: I was thinking about a different design but I can try to make a PR about it. This is fine as it is now

cfkanesan (approved):

PR #77, opened 2023-09-01
Add tdelta time operator

cfkanesan:
## Purpose

Add the weighted time difference operator.

## Code changes

- Add the `tdelta` function to the `time_operators` module.

### DIFF
--- a/src/idpi/operators/time_operators.py
+++ b/src/idpi/operators/time_operators.py
@@ -88,6 +88,36 @@ def delta(field: xr.DataArray, dtime: np.timedelta64) -> xr.DataArray:
     return result
 
 
+def tdelta(field: xr.DataArray, dtime: np.timedelta64) -> xr.DataArray:
+    """Compute weighted difference for a given delta in time.
+
+    Parameters
+    ----------
+    field : xr.DataArray
+        Field that contains the input data.
+    dtime : np.timedelta64
+        Time delta for which to evaluate the difference.
+
+    Raises
+    ------
+    ValueError
+        if dtime is not multiple of the field time step
+        or if the time step is not regular.
+
+    Returns
+    -------
+    xr.DataArray
+        The field weighted difference for the given time delta.
+
+    """
+    nsteps = get_nsteps(field.valid_time, dtime)
+    weights = (field.valid_time - field.ref_time) / dtime
+    weighted = field * weights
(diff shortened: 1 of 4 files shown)


cfkanesan:
The difference between the outputs seems to be discrete but there's no clear correlation to one of the input fields. I think it might just be due to the single precision. I'll increase the tolerance on the test.

image

src/idpi/operators/time_operators.py
```diff
@@ -88,6 +88,36 @@ def delta(field: xr.DataArray, dtime: np.timedelta64) -> xr.DataArray:
     return result
 
 
+def tdelta(field: xr.DataArray, dtime: np.timedelta64) -> xr.DataArray:
+    """Compute weighted difference for a given delta in time.
```
cosunae, line 92: could we improve here the name of the operator and documentation so that it reflects better what it does? 
Is this computing an average rate ?

cfkanesan: Added a paragraph in the docstring

cosunae: thinking again how we could name this operator. Isnt this a moving average difference in time, with a sampling rate defined by dtime ? If so we could rename as something like `tdiff_moving_average` ?

cfkanesan: It relies on the input being a time averaged value from the start of the computation. I think `tdiff_moving_average` is a bit misleading because it suggests that the operator is performing an averaging aggregation. Maybe `resample_average`? We loose however the correspondence with fieldextra.

cosunae: yes, I this is right. I think `resample_average` is a better name. I think is find to loose the mapping with fieldextra, at some point in time the mapping with fieldextra wont be necessary and it is better to have names that describe well the mathematical operation than applying historical reasons why it was named like that

cfkanesan: done

tests/test_idpi/test_time_ops.py
```diff
@@ -47,6 +47,40 @@ def test_delta(data_dir, fieldextra):
     assert_allclose(observed, expected.transpose("epsd_1", "time", ...))
 
 
+def test_tdelta(data_dir, fieldextra):
```
cosunae, line 50: rename also this test accordingly?

PR #79, opened 2023-09-11
Support gis

cfkanesan:
## Purpose

Add `support_gis` module including some functions to work with geographic coordinates.

### DIFF
--- a/src/idpi/operators/gis.py
+++ b/src/idpi/operators/gis.py
@@ -211,12 +211,22 @@ def vref_rot2geolatlon(
         x and y components of the vector field w.r.t. the geo lat lon coords.
 
     """
-    if u.origin["x"] != 0.0 or v.origin["y"] != 0.0:
+    valid_origin = {d: 0.0 for d in tuple("xyz")}
+    if u.origin != valid_origin or v.origin != valid_origin:
         raise ValueError("The vector fields must be destaggered.")
 
     grid = get_grid(u.geography)
     lon, lat = rot2geolatlon(grid)
+    return _vref_rot2geolatlon(u, v, lon, lat, grid)
 
+
+def _vref_rot2geolatlon(
+    u: xr.DataArray,
+    v: xr.DataArray,
+    lon: xr.DataArray,
+    lat: xr.DataArray,
+    grid: RotLatLonGrid,
+) -> tuple[xr.DataArray, xr.DataArray]:
     deg2rad = np.pi / 180
     sin_np = np.sin(deg2rad * grid.north_pole_lat)
     cos_np = np.cos(deg2rad * grid.north_pole_lat)
--- a/src/idpi/operators/gis.py
+++ b/src/idpi/operators/gis.py
@@ -1,4 +1,4 @@
-"""Support geospatial logic."""
+"""Geospatial logic."""
 
 # Standard library
 import dataclasses as dc
--- a/src/idpi/operators/gis.py
+++ b/src/idpi/operators/gis.py
@@ -126,17 +126,17 @@ def rot2geolatlon(grid: RotLatLonGrid) -> tuple[xr.DataArray, xr.DataArray]:
     arg1 = cos_np_lat * cos_lat * cos_lon + sin_np_lat * sin_lat
     lat = rad2deg * np.arcsin(arg1)
     # ... compute longitude
-    arg1 = (
+    arg2 = (
         sin_np_lon * (-sin_np_lat * cos_lon * cos_lat + cos_np_lat * sin_lat)
         - cos_np_lon * sin_lon * cos_lat
     )
-    arg2 = (
+    arg3 = (
         cos_np_lon * (-sin_np_lat * cos_lon * cos_lat + cos_np_lat * sin_lat)
         + sin_np_lon * sin_lon * cos_lat
     )
     # BUG: changes sign when arg2 is negative and less than threshold
-    arg2 = xr.where(np.abs(arg2) < 1e-20, 1e-20, arg2)
-    lon = rad2deg * np.arctan2(arg1, arg2) % 360
+    arg4 = xr.where(np.abs(arg3) < 1e-20, 1e-20, arg3)
+    lon = rad2deg * np.arctan2(arg2, arg4) % 360
 
     return lon, lat
 
--- a/src/idpi/operators/support_gis.py
+++ b/src/idpi/operators/support_gis.py
@@ -69,7 +69,6 @@ def get_grid(geo: Mapping[str, typing.Any]) -> RotLatLonGrid:
         object representing the rotated lat lon grid.
 
     """
-
     _check_requirements(geo)
 
     ni = geo["Ni"]
--- a/src/idpi/operators/support_gis.py
+++ b/src/idpi/operators/support_gis.py
@@ -155,6 +155,10 @@ def geolatlon2swiss(
     tuple[xarray.DataArray, xarray.DataArray]
         x and y coordinates in the Swiss LV03 coordinate system.
 
+    Notes
+    -----
+    Approximate formula published by swisstopo, precision in the order of 1 meter
+
     """
     norm_lat = ((lat * 3.6) - 169.02866) / 10
     lon = lon.where(lon < 180, lon - 360)
--- a/src/idpi/operators/support_gis.py
+++ b/src/idpi/operators/support_gis.py
@@ -47,11 +47,14 @@ def _check_requirements(geo: Mapping[str, typing.Any]) -> None:
     errors = {key: geo[key] for key in requirements if requirements[key] != geo[key]}
     if errors:
         msg = f"Unsupported values for keys: {errors}"
-        raise RuntimeError(msg)
+        raise ValueError(msg)
 
 
 def get_grid(geo: Mapping[str, typing.Any]) -> RotLatLonGrid:
     """Get grid parameters for a given field.
+
+    Only fields defined on regular grids in rotlatlon coordinates,
+    without rotation nor flipped axes are supported.
 
     Parameters
     ----------
@@ -61,7 +64,7 @@ def get_grid(geo: Mapping[str, typing.Any]) -> RotLatLonGrid:
     Raises
     ------
     ValueError
-        if the field is not defined on a rotated lat lon grid.
+        if the field does not fulfill the conditions above.
 
     Returns
     -------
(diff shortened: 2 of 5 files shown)


src/idpi/operators/support_gis.py
```diff
@@ -0,0 +1,229 @@
...
+        y component of the vector field w.r.t. a rotated lat lon grid.
+
+    Returns
+    -------
+    tuple[xarray.DataArray, xarray.DataArray]
+        x and y components of the vector field w.r.t. the geo lat lon coords.
+
+    """
+    if u.origin["x"] != 0.0 or v.origin["y"] != 0.0:
+        raise ValueError("The velocity fields must be destaggered.")
```
victoria-cherkas, line 211: I guess other fields other than velocities can be vectors? I'd maybe change valueerror to 
```suggestion
        raise ValueError("The vector fields must be destaggered.")
```

src/idpi/operators/support_gis.py
```diff
@@ -0,0 +1,229 @@
...
+
+    Parameters
+    ----------
+    geo : Mapping[str, Any]
+        Grib keys related to the geography of the field.
+
+    Raises
+    ------
+    ValueError
+        if the field is not defined on a rotated lat lon grid.
```
victoria-cherkas, line 64: I see that a RuntimeError is raised via _check_requirements but no ValueError?

src/idpi/operators/support_gis.py
```diff
@@ -0,0 +1,229 @@
...
+        "jScansPositively": 1,
+    }
+    errors = {key: geo[key] for key in requirements if requirements[key] != geo[key]}
+    if errors:
+        msg = f"Unsupported values for keys: {errors}"
+        raise RuntimeError(msg)
+
+
+def get_grid(geo: Mapping[str, typing.Any]) -> RotLatLonGrid:
+    """Get grid parameters for a given field.
```
victoria-cherkas, line 54: Docstring should maybe state that this function gets grid parameters for a given field _defined on rot lat lon grid_, since it raises error otherwise?

src/idpi/operators/support_gis.py
```diff
@@ -0,0 +1,229 @@
...
+    # ... compute longitude
+    arg1 = (
+        sin_np_lon * (-sin_np_lat * cos_lon * cos_lat + cos_np_lat * sin_lat)
+        - cos_np_lon * sin_lon * cos_lat
+    )
+    arg2 = (
+        cos_np_lon * (-sin_np_lat * cos_lon * cos_lat + cos_np_lat * sin_lat)
+        + sin_np_lon * sin_lon * cos_lat
+    )
+    arg2 = xr.where(np.abs(arg2) < 1e-20, 1e-20, arg2)  # BUG: changes sign?
```
victoria-cherkas, line 134: arg2 changes sign always or only when 1e-20? Think this comment should be expanded slightly.

src/idpi/operators/support_gis.py
victoria-cherkas, line 1: small thing but does this file need the word support? gis.py is more concise and 'support' seems superfluous

victoria-cherkas (commented):
Just a couple of comments, otherwise looks good to me!

src/idpi/operators/support_gis.py
cfkanesan: Good point

src/idpi/operators/support_gis.py
cfkanesan: When arg2 is negative but the absolute value is less than 1e-20, it is replaced by a value that is positive.

src/idpi/operators/support_gis.py
cfkanesan: I'm just reusing the fieldextra naming here. It can be removed if that brings little value.

victoria-cherkas: Yeah i think should be dropped. All the operator modules support something I guess

src/idpi/operators/support_gis.py
victoria-cherkas: Ah ok changes sign but not == 1e-20 as expected?

cfkanesan: I expanded the comment. Does it make more sense now?

victoria-cherkas: Sorry my question was whether the value/modulus changes as well as the sign or not.

victoria-cherkas: Because the expected sign is positive, expected value is 1e-20 ? Or do i misunderstand?

cfkanesan: I think it's just a guard to avoid a division by zero error. Arctan of +inf is 0.5 pi and -inf is -0.5 pi. The modulus does not catch the sign change. This probably went undetected because the range of values for which it misbehaves is tiny.

victoria-cherkas: I understand the need, I was asking about the value and not the sign, but nevermind.

victoria-cherkas (approved):

tests/test_idpi/test_gis.py
```diff
@@ -0,0 +1,55 @@
...
+
+    reader = grib_decoder.GribReader([datafile], ref_param="T")
+    ds = reader.load_cosmo_data(["U_10M", "V_10M"])
+
+    u_g, v_g = gis.vref_rot2geolatlon(ds["U_10M"], ds["V_10M"])
+
+    fx_ds = fieldextra("n2geog")
+
+    assert_allclose(u_g.isel(z=0), fx_ds["U_10M"], atol=1e-5, rtol=1e-6)
+    assert_allclose(v_g.isel(z=0), fx_ds["V_10M"], atol=1e-5, rtol=1e-6)
```
twicki, line 55: in this filewe combine self-contained tests with tests that have fieldextra and file-system dependencies. Is the idea to bunch them together because they test the same module or do we want to separate self-contained tests from tests that depend on other code?

src/idpi/operators/gis.py
```diff
@@ -0,0 +1,233 @@
...
+    # ... cache trigonometric operations
+    sin_lat = np.sin(deg2rad * grid.rlat)
+    cos_lat = np.cos(deg2rad * grid.rlat)
+    sin_lon = np.sin(norm_lon)
+    cos_lon = np.cos(norm_lon)
+    # ... compute latitude
+    arg1 = cos_np_lat * cos_lat * cos_lon + sin_np_lat * sin_lat
+    lat = rad2deg * np.arcsin(arg1)
+    # ... compute longitude
+    arg1 = (
```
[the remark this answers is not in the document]
cfkanesan: No it's the same way it was done in fieldextra but it does have some benefits in terms of memory footprint

src/idpi/operators/gis.py
```diff
@@ -0,0 +1,233 @@
...
+    v : xarray.DataArray
+        y component of the vector field w.r.t. a rotated lat lon grid.
+
+    Returns
+    -------
+    tuple[xarray.DataArray, xarray.DataArray]
+        x and y components of the vector field w.r.t. the geo lat lon coords.
+
+    """
+    if u.origin["x"] != 0.0 or v.origin["y"] != 0.0:
```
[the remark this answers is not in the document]
cfkanesan: Good question. Should we expect fields to be staggered in different dimensions than the component that they contain?

src/idpi/operators/gis.py
```diff
@@ -0,0 +1,233 @@
...
+        raise ValueError("The vector fields must be destaggered.")
+
+    grid = get_grid(u.geography)
+    lon, lat = rot2geolatlon(grid)
+
+    deg2rad = np.pi / 180
```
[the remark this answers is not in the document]
cfkanesan: I don't understand the suggestion, sorry.

src/idpi/operators/gis.py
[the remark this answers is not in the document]
twicki: I would separate the two here:
- I think in increases readability
- Optimizing memory footprint and shutting the door on potential parallelization seems like premature optimization to me
If we are very certain we want to go the route of minimizing memory footprint because we need to we can go that route though.

tests/test_idpi/test_gis.py
cfkanesan: You are suggesting to introduce a separation between unit and integration tests? I think it's good to have the tests organised in a similar way to the source code.

src/idpi/operators/gis.py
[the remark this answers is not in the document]
twicki: unclear - I think the E-grid stagges both ways, but I'm not sure if we're using it. This approach could work because we're saying they are specific components of a vector field.

src/idpi/operators/gis.py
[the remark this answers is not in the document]
twicki: I would take the block of https://github.com/MeteoSwiss-APN/icon_data_processing_incubator/blob/support-gis/src/idpi/operators/gis.py#L220:L231 and move it to its own function to keep the level of complexity of the code within this function (`grid = get_grid(u.geography)` and `lon, lat = rot2geolatlon(grid)`) on the same level. This does not change anything functionally though

src/idpi/operators/gis.py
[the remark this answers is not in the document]
cfkanesan: Can you be specific about the change that you are requesting and what potential parallelisation we are closing the door on?

tests/test_idpi/test_gis.py
twicki: I would split unit and integration tests - mostly to have multiple entrypoints for different levels of testing. But achieving this could also be done by marking the tests if you think the testfolder structure matching files 1:1 would be suited better here. 

Do we have this pattern in many other places as well?

src/idpi/operators/gis.py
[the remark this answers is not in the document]
twicki:
```suggestion
    arg3 = (
```
If we keep naming it arg1 and are writing into the same temporary memory there is a dependency on the lat computation happening before the lon computation which is not necessary here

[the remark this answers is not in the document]
cfkanesan: okay but that's not actually how it works in python. The reassignment to arg1 does not reuse the memory space previously allocated by the object that was labelled as arg1. It assigns the arg1 label to a new object in a new memory space. The reference counter for the previously labelled arg1 object then drops to zero and the memory is freed. That's the benefit that I was referring to: the reduction of the peak memory consumption. To enable parallelisation, this operator would need to be split into multiple functions that can be dispatched so the naming of the local variables doesn't really matter. However, I do prefer the immutable style.

tests/test_idpi/test_gis.py
cfkanesan: Right now we're not making any distinction and it's not really in the scope of this PR to address this.

twicki (approved):
LGMT

PR #83, opened 2023-09-28
Regrid operator

cfkanesan:
## Purpose

Add the regridding operator.

## Code changes:

- Add `regrid` module.

## Requirements changes:

- Add dependency on rasterio

## Notes

- Although the operator will broadcast the eps, time and z dimensions, the test case only includes a single level field.
- The tested area is restricted to avoid including undefined values due to a small difference between fieldextra and rasterio.


src/idpi/operators/regrid.py
```diff
@@ -0,0 +1,184 @@
...
+
+    return CRS.from_string(
+        f"+proj=ob_tran +o_proj=longlat +o_lat_p={lat} +lon_0={lon} +datum=WGS84"
+    )
+
+
+def _relimit(longitude: float) -> float:
+    if longitude > 180:
+        return longitude - 360
+    return longitude
```
dominichofer, line 41: Is this an incomplete renormalization of arbitrary angles to `(-pi, +pi]`?
Or is this renormalizing angles in `[0, 2*pi)` to `[-pi, +pi)`?
This function needs documentation on pre and post conditions, which explicitly mentions which bounds (upper/lower) are included.

cfkanesan: Because the domain is centered on the origin the xmin is usually negative but sometimes it's given in the [0, 360] range. In that case, xmin > xmax and some of the formulas breakdown. This function is supposed to make sure that xmin and xmax are compatible.

dominichofer: I understand, and this would replace the helper function with a general purpose function, that normalizes any angle to the interval [-180, +180). If you like it, add it, if you don't, ignore it.
```
import math

def normalize_angle(angle: float) -> float:
    return math.fmod(angle + 180, 360) - 180
```

src/idpi/operators/regrid.py
```diff
@@ -0,0 +1,184 @@
...
+        op : str
+            fieldextra out_regrid_target definition
+            i.e. crs,xmin,ymin,xmay,ymax,dx,dy.
+
+        """
+        crs_str, *grid_params = op.split(",")
+        crs = CRS.from_string(CRS_ALIASES[crs_str])
+        xmin, ymin, xmax, ymax, dx, dy = map(float, grid_params)
+        nx = (xmax - xmin) / dx + 1
+        ny = (ymax - ymin) / dy + 1
```
dominichofer, line 116: if xmax == xmin, calling dx would cause a division by zero. But I think it's fair to assume this will never happen, right?

src/idpi/operators/regrid.py
```diff
@@ -0,0 +1,184 @@
...
+
+    @property
+    def dy(self) -> float:
+        return (self.ymax - self.ymin) / (self.ny - 1)
+
+    @property
+    def transform(self) -> transform.Affine:
+        return transform.from_origin(
+            west=self.xmin - self.dx / 2,
+            north=self.ymax + self.dy / 2,
```
dominichofer, line 133: why half a delta?

src/idpi/operators/regrid.py
```diff
@@ -0,0 +1,184 @@
...
+        )
+        return output[::-1]
+
+    return xr.apply_ufunc(
+        reproject_layer,
+        field,
+        input_core_dims=[["y", "x"]],
+        output_core_dims=[["y1", "x1"]],
+        vectorize=True,
+    ).rename({"x1": "x", "y1": "y"})
```
dominichofer, line 184: Why don't you `output_core_dims=[["y", "x"]],` and skip the renaming?
In case this is a workaround for xr.apply_ufunc's inability to have overlapping input and output core dims, maybe document that this is a workaround.

tests/test_idpi/fieldextra_templates/test_regrid.nl
```diff
@@ -0,0 +1,70 @@
...
+ n_ompthread_collect=1
+ n_ompthread_generate=1,1
+ out_cost_expensive=1000
+/
+&GlobalResource
+ dictionary="/project/s83c/fieldextra/tsa/resources/dictionary_cosmo.txt"
+ grib_definition_path="/project/s83c/fieldextra/tsa/resources/eccodes_definitions_cosmo",
+                        "/project/s83c/fieldextra/tsa/resources/eccodes_definitions_vendor"
+ grib2_sample="/project/s83c/fieldextra/tsa/resources/eccodes_samples/COSMO_GRIB2_default.tmpl"
+ rttov_coefs_path="/project/s83c/fieldextra/tsa/resources/rttov_coefficients"
```
dominichofer, line 19: There's no /project on Balfrin.
If this code needs to run on Balfrin eventually, we should look for a better solution. If not, fine with me.

tests/test_idpi/test_regrid.py
```diff
@@ -0,0 +1,26 @@
...
+    hzerocl = fhzerocl(ds["T"], ds["HHL"], extrapolate=True)
+    out_regrid_target = "swiss,549500,149500,650500,250500,1000,1000"
+    dst = regrid.RegularGrid.parse_regrid_operator(out_regrid_target)
+    hzerocl.attrs["geography"] = ds["HHL"].geography
+    observed = regrid.regrid(hzerocl, dst, regrid.Resampling.bilinear)
+
+    fx_ds = fieldextra("regrid", out_regrid_target=out_regrid_target)
+    expected = fx_ds["HZEROCL"]
+
+    assert_allclose(observed, expected, rtol=5e-4)
```
dominichofer, line 26: Why is this a relative error?
AFAIK zero-zero is within the map, which means the new grid can only be off by a few millimeters close to zero-zero, but meters further out. Naively I would assume the new grid may not be off more than a few meters everywhere.

tests/test_idpi/test_regrid.py
```diff
@@ -0,0 +1,26 @@
+# Third-party
+from numpy.testing import assert_allclose
+
+# First-party
+from idpi import grib_decoder
+from idpi.operators import regrid
+from idpi.operators.hzerocl import fhzerocl
+
+
+def test_regrid(data_dir, fieldextra):
```
dominichofer, line 10: Is this a regression test against fieldextra?
If so, maybe add it in a doc string.

dominichofer (commented):
I cannot judge the content of `tests/test_idpi/fieldextra_templates/test_regrid.nl` in a reasonable amount of time. If it's tested, this is fine, but I don't see where. Where is it used? Where is it tested?

src/idpi/operators/regrid.py
cfkanesan: Thanks for the suggestion, I'll use the version from numpy since that's readily available in the current module.

cfkanesan (commented):
Thanks for the review @dominichofer! I'll try to add some comments to hopefully make things clearer.

cfkanesan:
> I cannot judge the content of `tests/test_idpi/fieldextra_templates/test_regrid.nl` in a reasonable amount of time. If it's tested, this is fine, but I don't see where. Where is it used? Where is it tested?

It's used by the fieldextra fixture to drive fieldextra to generate the expected values for the regression test.

src/idpi/operators/regrid.py
dominichofer: even better

src/idpi/operators/regrid.py
dominichofer: I think it's okay to leave that little bit of uncertainty in the code. It seems to work and we can always improve it when we learn more about it. LGTM

tests/test_idpi/test_regrid.py
dominichofer: With that wider scope, I agree, such a comment is unnecessary.

dominichofer:
> > I cannot judge the content of `tests/test_idpi/fieldextra_templates/test_regrid.nl` in a reasonable amount of time. If it's tested, this is fine, but I don't see where. Where is it used? Where is it tested?
> 
> It's used by the fieldextra fixture to drive fieldextra to generate the expected values for the regression test.

Which means the test assures its correctness. So I think that file is fine as well.

dominichofer (approved):
LGTM

dominichofer (approved):

twicki (approved):
LGTM

PR #84, opened 2023-10-05
Config

cosunae:
## Purpose

the main intention of this PR is to simplify the logic in the grib_decoder by introducing a config object, which can be used to set options like enable_dask or the data_scope = "cosmo"¦"ifs"

## Code changes:

- new config class that holds configuration options
- simplify the API of the grib_decoder (removing function overloads since the grib_decoder API is not agnostic of ifs or cosmo
- Set the right definitions of eccodes from the config (instead of API)
- load_cosmo_data -> load_fields renaming
- remove the mapping field names from IFS -> COSMO. The reason was to simplify the loader again, thinking that this transformation of IFS->COSMO would belong to a different module that could convert datasets from ifs to cosmo naming. The only added value of this mapping is the unit convertions, which could be done explicitly in the product or in a module that convert units from ifs to cosmo

## Checklist

- [ ] New operators are properly tested

Additionally, if the PR updates the version of the package

- [ ] The new version is properly set
- [ ] A Tag will be created after the bump

## Review


src/idpi/config.py
```diff
@@ -0,0 +1,64 @@
+"""global configuration for idpi."""
+
+# Standard library
+from typing import Any, Literal
+
+config: dict = {}
+
+
+class set:
```
cfkanesan, line 9: The name currently clashes with the built-in set. In any case, I think that the name could be more explicit.

src/idpi/config.py
```diff
@@ -0,0 +1,64 @@
...
+    ----------
+    config : dict, optional
+        dictionary object use to hold the configuration.
+        Default will use the config object in this module
+    **kwargs :
+        the configuration key-value pairs to set.
+
+    """
+
+    def __init__(self, config: dict = config, **kwargs):
```
cfkanesan, line 22: Relying on the default capture of a mutable variable to share state is a bit an unusual practice.

src/idpi/config.py
```diff
@@ -0,0 +1,64 @@
...
+        self._config = config
+        self._record: list[tuple[Literal["insert", "replace"], str, Any]] = []
+
+        if kwargs:
```
cfkanesan, line 26: This conditional check is redundant

src/idpi/config.py
```diff
@@ -0,0 +1,64 @@
...
+            for key, value in kwargs.items():
+                if key in self._config:
+                    self._record.append(("replace", key, self._config[key]))
+                else:
+                    self._record.append(("insert", key, None))
+
+                self._config[key] = value
+
+    def __enter__(self) -> dict:
+        return self._config
```
cfkanesan, line 36: I don't see the use case for returning the updated config since the global/passed config is mutated in place. More importantly, the actual mutation should happen in the enter function and not in the init. Maybe this calls for the context manager to be implemented using the function approach?

src/idpi/grib_decoder.py
```diff
@@ -32,8 +31,6 @@
     "isobaricInPa": ("pressure", 0.0),
     "surface": ("surface", 0.0),
 }
-_ifs_allowed = True
-_cosmo_allowed = True
```
cfkanesan, line 36: I don't think this removes the need to exclusively read cosmo or ifs data. The mapping is no longer applied but the actual issue was in the stale eccodes definitions cache that can not be cleared.

src/idpi/tasking.py
```diff
@@ -0,0 +1,18 @@
+"""functionality for tasking and parallel computing."""
```
cfkanesan, line 1: Perhaps the implementation of this module should be moved to PR #86

src/idpi/config.py
cosunae: right

src/idpi/config.py
cosunae: I will change it to set_values

src/idpi/config.py
cosunae: Returning a global config is an internal detail, and not part of the contract. No one is accessing directly the global object, the internal implementation could be creating clones of it. 
Why should the actual mutation happen in the init?

src/idpi/grib_decoder.py
cosunae: right, the config does not remove the need for this. However, this protection has also its pitfalls in my opinion. One needs to set those variables properly, which is error prone. 
There are several places with 
` idpi.config.get("data_scope", "cosmo") == "cosmo"`
that required setting the variables. 
If someones adds a new function that loads additional fields or require entering the scope of cosmo, ifs data, they should not forget that this variables must be set. 
Trying to avoid that, before we had wrapper functions duplicated for ifs and cosmo that are just there so that these variables are not forgotten. Which pollutes considerably the API. 
I think is a trade off, but provided that mixing IFS data and COSMO data in our processing is probably not a use case at this point in time, I thought the solution does not pay off. 
At some point, some projects might need to combine the processing of various data sources, but then I would argue that we should invest in fixing the issue with eccodes, rather than trying to protect our codebase. I do not see any fundamental reason why we could not clear the cache of eccodes, it is probably just a bug. 
Thoughts on this?

src/idpi/config.py
cfkanesan: > Why should the actual mutation happen in the init?

Because the context manager is implemented using a class, it's actually not forbidden to dissociate the instantiation from the entering of the context e.g.
```
mgr = set(foo="bar")
...
with mgr:
    pass
```
The mutation would be expected to happen in the with block. However, with the current implementation, it would happen as soon as the init function is called.

cosunae: I see, thanks for the clarification. I am changing this to use a contextmanager decorator

src/idpi/tasking.py
cosunae: yes, that was a mistake. It belongs to #86

tests/test_idpi/test_config.py
```diff
@@ -0,0 +1,15 @@
+# First-party
+import idpi.config
+
+
+def test_config():
+    with idpi.config.set_values(opt1=1):
+        assert idpi.config.get("opt1") == 1
+
+        with idpi.config.set_values(opt1=False, opt2=2):
+            assert not idpi.config.get("opt1")
```
cfkanesan, line 10:
```suggestion
            assert idpi.config.get("opt1") is False
```
not evals to True for anything falsey such as an empty container

tests/test_idpi/test_config.py
```diff
@@ -0,0 +1,15 @@
...
+            assert idpi.config.get("opt2") == 2
+
+        assert idpi.config.get("opt1") == 1
```
cfkanesan, line 13: probably the `opt2` key should be not set at this point?

cfkanesan (approved):
Minor comments on the tests but looks good otherwise

tests/test_idpi/test_config.py
cosunae: done

PR #86, opened 2023-10-05
Config dask

cosunae:
## Purpose
use the config to set the property "enable_dask" to avoid having to pass it as args to various functions and avoid inconsistencies.

## Code Changes

* Get the property enable_dask via the config.
* Hide dask behind a tasking module, so that dask is not a dependency of any of the products. This way we adopt a task parallel model that can be configure with any tasking runtime system.


src/idpi/grib_decoder.py
```diff
@@ -5,18 +5,17 @@
...
-import dask
 import earthkit.data  # type: ignore
 import eccodes  # type: ignore
 import numpy as np
 import xarray as xr
 
 # First-party
 import idpi.config
+import idpi.tasking as tasking
 from idpi.product import ProductDescriptor
```
cfkanesan, line 19:
```suggestion
from . import tasking
from .product import ProductDescriptor
```

PR #90, opened 2023-10-25
Minor cleanup

twicki:
Minor cleanup on the data-path fixtures

### DIFF
--- a/tests/test_idpi/conftest.py
+++ b/tests/test_idpi/conftest.py
@@ -2,6 +2,7 @@
 # Standard library
 import subprocess
 from pathlib import Path
+from typing import Callable, List
 
 # Third-party
 import pytest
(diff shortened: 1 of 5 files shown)


tests/test_idpi/test_lat_ops.py
```diff
@@ -12,14 +14,12 @@
 from idpi.operators.hzerocl import fhzerocl
 
 
-@pytest.fixture
-def data_dir():
```
cfkanesan, line 16: The purpose of this fixture was to override the top-level data_dir fixture that the fieldextra fixture was using to define the default paths. With this change, you need to pass in updated paths to the fieldextra call.

tests/test_idpi/conftest.py
```diff
@@ -2,21 +2,50 @@
...
 import pytest
 import xarray as xr
 from jinja2 import Environment, FileSystemLoader
 
 
-@pytest.fixture
-def data_dir():
-    """Base data dir."""
+@pytest.fixture(name="data_dir")
+def data_dir_fixture():
```
cfkanesan, line 14: What is the purpose of this change?

tests/test_idpi/conftest.py
```diff
@@ -2,21 +2,50 @@
...
+    """Provide the path to the cropped dataset for testing.
+
+    Returns:
+        Path: Path to the dataset.
```
cfkanesan, line 18: Could you use the numpy doc format?

tests/test_idpi/conftest.py
```diff
@@ -32,7 +61,9 @@ def template_env():
 
 
 @pytest.fixture
-def fieldextra(tmp_path, data_dir, template_env, fieldextra_executable):
+def fieldextra(
+    tmp_path, data_dir, template_env, fieldextra_executable
+) -> Callable[..., xr.Dataset | List[xr.Dataset]]:
```
cfkanesan, line 66:
```suggestion
) -> Callable[..., xr.Dataset | list[xr.Dataset]]:
```
the built in list is now generic

cfkanesan (changes requested):

PR #92, opened 2023-11-16
Add data_source module

cfkanesan:
## Purpose

The goal is extract the logic regarding the fetching of data from the construction of the xarray dataset. The DataSource class takes over the responsibility of:
- loading the appropriate GRIB definitions based on the `data_scope` with a fallback on the request model field
- dispatching the request to either given datafiles or an FDB instance
- injecting contextual values into the request

Driving the data loading based on the product descriptions will be handled in a future PR.

## Code changes:

- Add `data_source` module 
- **Breaking:** removed the `load_fields` method from the GribReader class
- **Breaking:** the `load` method of GribReader class now takes a mapping of labels to requests and returns a mapping of labels to `xarray.DataArray`


victoria-cherkas (approved):
Looks great! (minus pre commit fail)

PR #97, opened 2023-12-04
Add run_products function

cfkanesan:
## Purpose

Added a function that manages the running of multiple products together. Notably, the functions makes sure that redundant requests are only submitted once to the data source.

## Code changes

- Added the `idpi.product.run_products` function.

### DIFF
--- a/src/idpi/product.py
+++ b/src/idpi/product.py
@@ -2,21 +2,22 @@
 # Standard library
 import dataclasses as dc
 from abc import ABCMeta, abstractmethod
+from collections import Counter
 from typing import NamedTuple
 
 # Local
-from . import tasking
+from . import data_source, grib_decoder, tasking
 
 
 class Request(NamedTuple):
     param: str
-    levtype: str | None = None
+    levtype: str = "ml"
     levelist: tuple[int, ...] | None = None
 
 
 @dc.dataclass
 class ProductDescriptor:
-    input_fields: list[Request]
+    input_fields: dict[str, Request]
 
 
 class Product(metaclass=ABCMeta):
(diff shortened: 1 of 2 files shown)


src/idpi/product.py
```diff
@@ -43,3 +44,45 @@ def __call__(self, *args, **kwargs):
...
+    for product in products:
+        for name, req in product.descriptor.input_fields.items():
+            name_set = req_labels.setdefault(req, set())
+            name_set.add(name)
+
+    # check for any conflicts in the names
+    cnt = Counter(name for names in req_labels.values() for name in names)
+    conflicts = [name for name, count in cnt.items() if count > 1]
+    if conflicts:
+        raise RuntimeError(f"These names are involved in conflicts: {conflicts}")
```
victoria-cherkas, line 62: Its not immediately clear what the context of this conflict check is for, can you add some more info to the comment? Also is name representing param here?

src/idpi/product.py
```diff
@@ -43,3 +44,45 @@ def __call__(self, *args, **kwargs):
     @property
     def descriptor(self):
         return self._desc
+
+
+def _merge_requests(
```
victoria-cherkas, line 49: Can you document this function, and why do we need aliases?

src/idpi/product.py
cfkanesan: We are merging mappings of names to requests, it's okay if there are multiple names for a single request but not if there are multiple requests for a single name. The name is indeed used as a key in the output dataset. The name references a field request and the corresponding `xarray.DataArray` returned by the data source.

milosbeliczuhlke (approved):

victoria-cherkas (approved):
Makes sense, thanks for adding the docs 👍

PR #100, opened 2023-12-11
Reduced test data

cfkanesan:
## Purpose

Run tests on a reduced test dataset by default to speed up the test suite.

## Code changes:

- Added the custom pytest marker `data` 
- Remove the `system_defintion` module

Superseeds #88


tests/test_idpi/conftest.py
```diff
@@ -27,12 +26,25 @@ def machine():
...
-        return Path("/project/s83c/rz+/icon_data_processing_incubator/data/SWISS")
-    raise RuntimeError("panic")
+    if machine != "tsa":
+        return None
+    base_dir = Path("/project/s83c/rz+/icon_data_processing_incubator/")
+    marker = request.node.get_closest_marker("data")
+    if marker is None:
+        return base_dir / "datasets/32_39x45_51"
+    match marker.args[0]:
+        case "original":
```
sfriedlizuehlke, line 39: Use an enum insted of plain strings?

cosunae (approved):
maybe have you considered the comment from Stefan?
Otherwise looks good to me

cfkanesan:
> maybe have you considered the comment from Stefan? Otherwise looks good to me

Yes but I'm not sure where to place the enum in order to make it useful for someone who is writing a test. There is a concept of a global namespace in pytest but I don't think that's picked up by vscode.

PR #107, opened 2023-12-15
Write to grib file

cfkanesan:
## Purpose

Provide the ability to write xarray.DataArray objects back to the GRIB format.

## Code changes

- Add `save` function to `grib_decoder` module

Notes:
- most operators currently do not update the metadata 
- the earthkit data metadata object is added to the DataArray attributes and contain one entire GRIB message

### DIFF
--- a/src/idpi/grib_decoder.py
+++ b/src/idpi/grib_decoder.py
@@ -330,6 +330,25 @@ def _get_type_of_level(field):
 
 
 def save(field: xr.DataArray, file_handle: io.BufferedWriter):
+    """Write field to file in GRIB format.
+
+    Parameters
+    ----------
+    field : xarray.DataArray
+        Field to write into the output file.
+    file_handle : io.BufferedWriter
+        File handle for the output file.
+
+    Raises
+    ------
+    ValueError
+        If the field does not have a metadata attribute.
+
+    """
+    if not hasattr(field, "metadata"):
+        msg = "The metadata attribute is required to write to the GRIB format."
+        raise ValueError(msg)
+
     idx = {
         dim: field.coords[key]
         for key in field.dims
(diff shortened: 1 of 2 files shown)


tests/test_idpi/test_grib_decoder.py
```diff
@@ -0,0 +1,24 @@
+# Third-party
+import xarray as xr
+
+# First-party
+from idpi import grib_decoder
+
+
+def test_save(data_dir, tmp_path):
```
petrabaumann, line 8: Wouldn't we need to add tests for all types of supported metadata?

petrabaumann (commented):
I know that I have not looked at the code for a long time. So I just have some question about exception handling and extension of unit tests.

src/idpi/grib_decoder.py
```diff
@@ -309,3 +314,36 @@ def load_fieldnames(
...
+    if field.vcoord_type == "model_level":
+        if field.origin["z"] == 0.0:
+            return "generalVerticalLayer"
+        elif field.origin["z"] == -0.5:
+            return "generalVertical"
+        else:
+            raise ValueError(f"Unsupported field origin in z: {field.origin['z']}")
+    else:
+        mapping = {vc: name for name, (vc, _) in VCOORD_TYPE.items()}
+        return mapping.get(field.vcoord_type, field.vcoord_type)
```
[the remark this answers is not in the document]
cfkanesan: The second argument of `mapping.get` is a default value. Basically, the idea is to use the value of `field.vcoord_type` as the `typeOfLevel` in case there are no entries in the `VCOORD_TYPE` dict for it.

tests/test_idpi/test_grib_decoder.py
cfkanesan: Yes, indeed, the coverage of this test is pretty minimal. I wanted to get some feedback on the general approach before starting to write a battery of tests.

sfriedlizuehlke (approved):

PR #110, opened 2023-12-22
Metadata override

cfkanesan:
## Purpose

Provide the functionality to update the attached GRIB message with relevant metadata

## Code changes

- Added `idpi.metadata.override` function

### DIFF
--- a/src/idpi/metadata.py
+++ b/src/idpi/metadata.py
@@ -1,12 +1,30 @@
+"""Manage GRIB metadata."""
+
 # Standard library
 import io
+import typing
 
 # Third-party
 import earthkit.data as ekd
 from earthkit.data.writers import write
 
 
-def override(message: bytes, **kwargs):
+def override(message: bytes, **kwargs: typing.Any) -> dict[str, typing.Any]:
+    """Override GRIB metadata contained in message.
+
+    Parameters
+    ----------
+    message : bytes
+        Byte string of the input GRIB message
+    kwargs : Any
+        Keyword arguments forwarded to earthkit-data GribMetadata override method
+
+    Returns
+    -------
+    dict[str, Any]
+        Updated message byte string along with the geography and parameter namespaces
+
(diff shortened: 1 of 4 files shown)


src/idpi/metadata.py
```diff
@@ -0,0 +1,39 @@
...
+        Keyword arguments forwarded to earthkit-data GribMetadata override method
+
+    Returns
+    -------
+    dict[str, Any]
+        Updated message byte string along with the geography and parameter namespaces
+
+    """
+    stream = io.BytesIO(message)
+    [grib_field] = ekd.from_source("stream", stream)
```
petrabaumann, line 29: I think that this works, as long as you do not have to use another product definition template to code the result.

petrabaumann (commented):
If I interpret the code correctly, you create a new GRIB2 message by creating it from that of the parent, i.e. using the parent's GRIB2 record as a sample. This mechanism will not work, if you have to use a new product definition template to code the result of an operation, e.g., computing the ensemble mean from member data.

cfkanesan:
> If I interpret the code correctly, you create a new GRIB2 message by creating it from that of the parent, i.e. using the parent's GRIB2 record as a sample. This mechanism will not work, if you have to use a new product definition template to code the result of an operation, e.g., computing the ensemble mean from member data.

Couldn't you update the relevant keys to reflect the ensemble mean?

petrabaumann:
Depending on the product definition template number, different keys have to be set.

victoria-cherkas (approved):
Looks good to me. As @petrabaumann says it might be that you need to override metadata in a particular order eg product definition number first and then other keys that come with that template. But i think this can be implemented in another PR when we need to modify such metdata, eg statistics.

victoria-cherkas:
If you leave that for another PR, perhaps you can add a comment to the override docstring explaining that the function is designed for modifications within the same template.

PR #111, opened 2023-12-28
Refactor constants

cfkanesan:
## Purpose

Improve the consistency of the physical constants

## Code changes 

- The constants module is renamed to physical constants and the `pc_` prefix is dropped in favour of aliasing the module to `pc` at the call sites


src/idpi/operators/omega_slope.py
```diff
@@ -30,6 +27,8 @@ def omega_slope(
     ps: xr.DataArray, etadot: xr.DataArray, ak: xr.DataArray, bk: xr.DataArray
 ):
     """Compute omega slope."""
+    surface_pressure_ref = 101325.0
+
```
victoria-cherkas, line 31: How come this is not in the physical_constants file?

cfkanesan: Seems more like reference value rather than a physical constant but I can move it there if you prefer.

victoria-cherkas: I think its better in the constant file, it coud be useful somewhere else too.

victoria-cherkas (approved):

PR #112, opened 2024-01-02
Add crop operator

cfkanesan:
## Purpose

Provide a cropping operator that maintains the metadata up to date. This is required when using an operator that relies on the metadata such the regridding to an automatically computed output grid. 

## Code changes:

- Added `idpi.operators.crop` module.


src/idpi/operators/crop.py
```diff
@@ -0,0 +1,63 @@
...
+    nj = ymax - ymin + 1
+
+    return xr.DataArray(
+        field.isel(x=slice(xmin, xmax + 1), y=slice(ymin, ymax + 1)),
+        attrs=metadata.override(
+            field.message,
+            longitudeOfFirstGridPointInDegrees=lon_min,
+            Ni=ni,
+            latitudeOfFirstGridPointInDegrees=lat_min,
+            Nj=nj,
```
victoria-cherkas, line 61: do you not also need to set latitudeOfLastGridPoint, longitudeOfLastGridPoint, numberOfDataPoints ? or are these keys derived?

victoria-cherkas: Also why do you set longitudeOfFirstGridPointInDegrees instead of longitudeOfFirstGridPoint ? Same question for latitude. longitudeOfFirstGridPoint/latitudeOfFirstGridPoint are the GRIB keys and +InDegrees are eccodes derived keys?

src/idpi/operators/crop.py
```diff
@@ -0,0 +1,63 @@
...
+
+# Local
+from .. import metadata
+from . import gis
+
+
+def crop(field: xr.DataArray, bounds: tuple[int, ...]) -> xr.DataArray:
+    """Crop the field to the given bounds.
+
+    Only fields defined on regular grids in rotlatlon coordinates,
```
petrabaumann, line 14: You could raise an error message if the gridDefinitionTemplateNumber differs from 0 or 1.

src/idpi/operators/crop.py
```diff
@@ -0,0 +1,63 @@
...
+    without rotation nor flipped axes are supported.
```
petrabaumann, line 15: I do not understand the remark "without rotation here".

src/idpi/operators/crop.py
petrabaumann: In fieldextra we set all the GRIB2 keys that are expected by the template. Lathough there may be some redundancy in the information. Perhaps one should at least check, if latitudeOfLastGridPoint and longitudeOfLastGridPoint are defined in the original message, and re-set them consistently. The keys latitude/longitudeOfFirst/LastGridPoint must be used, if the data is defined in micro degrees.

tests/test_idpi/test_crop.py
```diff
@@ -0,0 +1,25 @@
+# Third-party
+from numpy.testing import assert_equal
+
+# First-party
+from idpi.grib_decoder import GribReader
+from idpi.operators import crop
+
+
+def test_crop(data_dir):
```
petrabaumann, line 9: I would check the values of all keys related to the supported grid definition templates.

petrabaumann (commented):
I just had some add-ons to Victoria's comments.

src/idpi/operators/crop.py
petrabaumann: The GRIB2 standard recommends to code latitudes and longitudes in micro degrees. The basic angle of the production domain and subdivisions of it are be provided to manage cases where the recommended unit is not applicable to describe the extreme latitude and longitude values and the direction increments.

src/idpi/operators/crop.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    xmin, xmax, ymin, ymax = bounds
+
+    sizes = field.sizes
+    if (
+        xmin > xmax
+        or ymin > ymax
+        or any(v < 0 for v in bounds)
+        or xmax >= sizes["x"]
+        or ymax >= sizes["y"]
```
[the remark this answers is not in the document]
cfkanesan: I think that should be assured by the combination of these checks and the non negative ones on xmin, ymin

src/idpi/operators/crop.py
cfkanesan: > do you not also need to set latitudeOfLastGridPoint, longitudeOfLastGridPoint, numberOfDataPoints ? or are these keys derived?

I'm a bit unclear on what keys eccodes updates with the setValues function. Let me check whether the other keys updated.

src/idpi/operators/crop.py
cfkanesan: The rotated lat lon coordinate system can have an additional rotation around the vertical at the origin point.

src/idpi/operators/crop.py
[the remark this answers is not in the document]
victoria-cherkas: Ah yes I was forgetting these are indices rather than lat lon coords

src/idpi/operators/crop.py
cfkanesan: The message is contained within the field (as a byte string) but this key is not part of the geography namespace. It's possible to use earthkit data to decode the grib message however it seems a bit overkill.

tests/test_idpi/test_crop.py
cfkanesan: I'm checking the geography namespace as defined by earthkit data. If that's not sufficient I could decode the message to extract all of the relevant keys.

src/idpi/operators/crop.py
cfkanesan: updated to set the first and last points in microdegrees

src/idpi/operators/crop.py
petrabaumann: OK!

src/idpi/operators/support_operators.py
```diff
@@ -122,4 +122,5 @@ def get_grid_coords(n: int, x0: float, dx: float, dim: str) -> xr.DataArray:
         A 1-D field containing the coordinates of the grid along the given dimension
 
     """
-    return xr.DataArray(np.arange(n, dtype=np.float32) * dx + x0, dims=dim)
+    values = np.arange(n) * dx + x0
+    return xr.DataArray(np.round(values, 6), dims=dim)
```
petrabaumann, line 126: Perhaps it would make sense to add a comment on why you round to 6 decimals in the description of the function.

tests/test_idpi/test_crop.py
petrabaumann: Since you also change the numberOfDataPoints in the crop operator, you should at least also check that key.

src/idpi/operators/crop.py
petrabaumann: It was just an idea. One has to see that in the context of the entire logical flow, and here I do not yet have an overview.

petrabaumann (commented):
Overall fine with your changes. I still have to comments on the unit test and on a hardcoded value that should be commented for better traceability.

tests/test_idpi/test_crop.py
cfkanesan: It's not part of the geography namespace so I need to decode the message to read it.

petrabaumann: OK, then let's stay with the test as it is now, and have a look at ECMWF's namespace concept later on.

petrabaumann (approved):
Fine for me now. Perhaps you could still add the comment on the precision of coordinates in support_operators.py.

PR #117, opened 2024-02-23
Allow to set bits_per_value

cfkanesan:
## Purpose

Expose the `bit_per_value` when writing back to grib. This option overrides the value in the GRIB metadata and is set to a default of 16 in the earthkit.data lib. 

## Code changes:

- Added `bits_per_value` option to the `grib_decoder.save` function


victoria-cherkas (approved):
great! thanks for finding the solution

PR #126, opened 2024-04-02
Add mch_model_data module

cfkanesan:
## Purpose 

Provide a more user friendly interface to get model data from Meteoswiss

## Code changes

- Add `idpi.mch_model_data` module

### DIFF
--- a/src/idpi/mch_model_data.py
+++ b/src/idpi/mch_model_data.py
@@ -63,7 +63,7 @@ def get_from_polytope(request: mars.Request) -> dict[str, xr.DataArray]:
     ------
     RuntimeError
         if the required environment variables for polytope are not set.
-        
+
     Returns
     -------
     dict[str, xarray.DataArray]
--- a/src/idpi/mch_model_data.py
+++ b/src/idpi/mch_model_data.py
@@ -1,6 +1,16 @@
+"""Get Meteoswiss model data from FDB or polytope."""
+
+# Standard library
+import logging
+import os
+
+# Third-party
 import xarray as xr
 
-from . import mars, data_source, grib_decoder
+# Local
+from . import data_source, grib_decoder, mars
+
+logger = logging.getLogger(__name__)
 
 
 def get_from_fdb(request: mars.Request) -> dict[str, xr.DataArray]:
@@ -13,6 +23,8 @@ def get_from_fdb(request: mars.Request) -> dict[str, xr.DataArray]:
 
     Raises
     ------
+    RuntimeError
+        if the required environment variables for FDB are not set.
     ValueError
         if the request has a feature attribute.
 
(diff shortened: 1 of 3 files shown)


src/idpi/mch_model_data.py
```diff
@@ -0,0 +1,82 @@
...
+            "Define 'POLYTOPE_ADDRESS', 'POLYTOPE_USERNAME' and 'POLYTOPE_PASSWORD'."
+        )
+        logger.exception(msg)
+        raise RuntimeError(msg)
+    if request.feature is not None:
+        collection = "mchgj"
+    else:
+        collection = "mch"
+    logger.info("Getting request %s from polytope collection %s", request, collection)
+    source = data_source.DataSource(polytope_collection=collection)
```
ninaburg, line 81: This is a side remark but in the description of `DataSource.retrieve()` [on this line](https://github.com/MeteoSwiss-APN/icon_data_processing_incubator/blob/main/src/idpi/data_source.py#L65), I think you forgot to mention that we can also use _polytope_ to define the DataSource (alongside datafiles and fdb). Let me know if I didn't understand it correctly.

src/idpi/mch_model_data.py
```diff
@@ -0,0 +1,82 @@
...
+# Third-party
+import xarray as xr
+
+# Local
+from . import data_source, grib_decoder, mars
+
+logger = logging.getLogger(__name__)
+
+
+def get_from_fdb(request: mars.Request) -> dict[str, xr.DataArray]:
```
ninaburg, line 16: Could the `request` also be of type `str | tuple | dict` as seen in other places in the code?

ninaburg (commented):
The changes look good, but I have some questions. Since I'm still getting familiar with the project, I will await a second opinion before approving to ensure everything aligns well with the project's setup and goals.

src/idpi/mch_model_data.py
cfkanesan: Here it's better to ensure a full request because all keys need to be specified when making requests to FDB or polytope. In the rest of the code, mainly in the tests, the request is done against files and a request template is used to complete the request for FDB but this approach is a bit complicated for this API.

src/idpi/mch_model_data.py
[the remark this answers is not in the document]
cfkanesan: No it's not tested in unit testing because polytope is an external service and it would make the test conditional on the availability of the service. It would be possible mock the service but the mock would need to have some relatively complex logic due to how the polytope client interacts with the server. That's something that will need to be kept in sync with the real polytope service to ensure that the test does its job. For now, we're testing this manually with the demo but in the longer term it would make sense to introduce something like a nightly integration test that runs periodically.

src/idpi/mch_model_data.py
```diff
@@ -0,0 +1,50 @@
...
+        if the request has a feature attribute.
+
+    Returns
+    -------
+    dict[str, xarray.DataArray]
+        Dataset containing the requested data.
+
+    """
+    if request.feature is not None:
+        raise ValueError("FDB does not support the feature attribute")
```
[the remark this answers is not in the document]
cfkanesan: No, it's the feature extraction that is limited to polytope/gribjump. It provides the ability to extract the timeseries in a limited region (e.g. a single point) but, if you're interested in the full field, you can pass multiple time steps to FDB.

ninaburg (commented):
It looks good now for me. And thank you for the answers to my questions.

sfriedlizuehlke (approved):

